PR #31, opened 2015-02-20
remove useless dll soversion

xantares:
as reported by M. Valle on the mailing list the 19/02/2015,
should only happen when BUILD_SHARED_LIBS is ON,
but it can be removed afterall


stevengj:
This is fine with me.  On Windows, I guess the filename has nothing to do with the so-version (unlike e.g. Linux), so we can leave the suffix as just `.dll`.

PR #32, opened 2015-02-20
build python bindings with cmake

xantares:


stevengj:
It looks like this requires swig on the user's system.  However, the tarball is distributed with the swig-generated `.cxx` files included so that users don't need swig.  (swig should only be needed for people hacking nlopt itself.)

xantares:
Ok, but I want to hack too :)

stevengj:
Looks good, thanks.

Issue #30, opened 2015-02-09
need support for soft feasibility constraints


jpritikin:
At least for local gradient-based methods, [OpenMx](http://openmx.psyc.virginia.edu/) needs support for soft feasibility constraints expressed by returning NaN for the fit. We use these in situations where the constraint is not expected to be active at the MLE. For example, a model may assume a positive definite covariance matrix and return NaN when the matrix is not positive definite. It is probably possible to add an inequality constraint to ensure positive definiteness, but any plausible MLE will be positive definite. Therefore, it is much easier to score such points as NaN.

Another example is when we estimate confidence intervals using a profile likelihood. The fit function when estimating the confidence interval includes a squared distance from a target log likelihood value. Here we use a soft constraint such that large square distances (>100) are scored as NaN. The distance at the MLE should be zero so this constraint is never active at a valid MLE.

Our proprietary optimizer, NPSOL, can handle non-linear constraints but it also knows that we use NaN to express soft constraints. When NPSOL evaluates the fit and gets a NaN, it halves the step size and tries again. If this continues to fail after a few retries then it gives up. One way to implement a similar kind of functionality using NLOPT is to return std::numeric_limits<double>::max(). However, NLOPT takes this fit at face value and it screws up the Hessian approximation (tested with SLSQP).

Is there some existing functionality to handle this kind of situation in NLOPT? Or are you willing to accept patches implementing it?

stevengj:
This will depend a lot on the specific algorithm.  Algorithms that are gradient-based (or which internally construct an approximate gradient, like COBLYA) will have a hard time with something like this—they need differentiability, not to mention continuity.   Algorithms which don't exploit continuity, like ISRES or DIRECT, could definitely be modified to handle this, but those algorithms tend to converge more slowly.

In your example of squared distance, the simplest thing would just to be to add a `distance² ≤ 100` inequality constraint; I don't see why you need NaNs for this.

The positive-definiteness constraint is a bit harder to implement, because eigenvalues are not generally differentiable.  (SDP solvers handle positive-definiteness constraints, of course, but they require specialized methods.)  In that case, couldn't you just as easily write your matrix as `A = B' * B` and then put your degrees of freedom in `B` rather than `A`?  That would guarantee a positive semidefinite `A` implicitly with no loss of generality.

jpritikin:
[...]

PR #34, opened 2015-03-19
soft feasibility constraints for slsqp

jpritikin:
I'm not sure how to run the standard tests, but I've tested this against our test collection in OpenMx. For context, see issue #30.

### DIFF
--- a/slsqp/slsqp.c
+++ b/slsqp/slsqp.c
@@ -2044,12 +2044,16 @@ L190:
     *mode = line == 1 ? -2 : 1;
     goto L330;
 L200:
-    if (h1 <= h3 / ten || line > 10) {
-	goto L240;
+    if (nlopt_isfinite(h1)) {
+	    if (h1 <= h3 / ten || line > 10) {
+		    goto L240;
+	    }
+	    /* Computing MAX */
+	    d__1 = h3 / (two * (h3 - h1));
+	    alpha = MAX2(d__1,alfmin);
+    } else {
+	    alpha = MAX2(alpha*.5,alfmin);
     }
-/* Computing MAX */
-    d__1 = h3 / (two * (h3 - h1));
-    alpha = MAX2(d__1,alfmin);
     goto L190;
 /*   EXACT LINESEARCH */
 L210:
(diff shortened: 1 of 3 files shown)


slsqp/slsqp.c
```diff
@@ -2044,12 +2044,16 @@ static void slsqpb_(int *m, int *meq, int *la, int *
     *mode = line == 1 ? -2 : 1;
     goto L330;
 L200:
-    if (h1 <= h3 / ten || line > 10) {
-	goto L240;
+    if (isfinite(h1)) {
```
stevengj: `isfinite` is a C99 function, and right now NLopt is not requiring C99.   So, probably we need a check (similar to the check for `isinf`) in `configure.ac` that defines `HAVE_ISFINITE` if `isfinite` is available,  a `nlopt_isfinite(x)` function in `api/general.c` that calls `isfinite` if it is available and otherwise uses its own implementation, and a declaration of `nlopt_isfinite` in `util/nlopt-util.h`

e.g. doing `#include <float.h>` and doing `nlopt_isfinite(double x) { return fabs(x) <= DBL_MAX; }` should work in any C89 compiler.

slsqp/slsqp.c
```diff
@@ -2491,80 +2496,64 @@ nlopt_result nlopt_slsqp(unsigned n, nlopt_func f, void *f_data,
...
+		   if (prev_mode == -2 && !want_grad) break; /* just evaluated this point */
 	      case -2:
 		   eval_f_and_grad:
+		      want_grad = 1;
+	  case 1:{ /* don't need grad unless we don't have it yet */
+		  double *newgrad = 0;
+		  double *newcgrad = 0;
+		  if (want_grad) {
+			  newgrad = grad;
+			  newcgrad = cgradtmp;
```
stevengj: Please use the same indentation style as the surrounding code (4 spaces)

slsqp/slsqp.c
```diff
@@ -2491,80 +2496,64 @@ nlopt_result nlopt_slsqp(unsigned n, nlopt_func f, void *f_data,
...
-			for (k = 0; k < h[i].m; ++k, ++ii) {
-			     infeasibility_cur = 
-				  MAX2(infeasibility_cur, fabs(c[ii]));
-			     feasible_cur = 
-				  feasible_cur && fabs(c[ii]) <= h[i].tol[k];
-			     for (j = 0; j < n; ++ j)
-				  cgrad[j*U(mpi1) + ii] = cgradtmp[k*n + j];
-			}
+		   if (isfinite(fcur)) {
+			   if (want_grad) {
```
stevengj: you can just unconditionally set `want_grad = 0` here

stevengj:
Looks pretty good modulo my comments above; thanks for going through the spaghetti code that is SLSQP (thanks to its Fortran reverse-calling heritage).

As long as we are checking for infinity, if `fcur` is `-Inf`, probably we should just return immediately and call this the minimum.

jpritikin:
Is this good enough?

stevengj:
Looks good, thanks!

PR #38, opened 2015-05-04
Back away from inequality constraints that evaluate to NaN

jpritikin:
This permits slsqp to deal with the interaction between inequality constraints and soft feasibility constraints.

### DIFF
--- a/src/algs/slsqp/slsqp.c
+++ b/src/algs/slsqp/slsqp.c
@@ -2052,8 +2052,11 @@ L190:
     *mode = line == 1 ? -2 : 1;
     goto L330;
 L200:
+    if (line > 10) {
+            goto L240;
+    }
     if (nlopt_isfinite(h1)) {
-	    if (h1 <= h3 / ten || line > 10) {
+	    if (h1 <= h3 / ten) {
 		    goto L240;
 	    }
 	    /* Computing MAX */
@@ -2092,7 +2095,12 @@ L220:
 	}
 /* Computing MAX */
 	d__1 = -c__[j];
-	t += mu[j] * MAX2(d__1,h1);
+	if (nlopt_isfinite(d__1)) {
+	  t += mu[j] * MAX2(d__1,h1);
+	} else {
+	  t = d__1;
+	  break;
+	}
 /* L230: */
     }
     h1 = t - t0;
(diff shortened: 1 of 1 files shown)


jpritikin:
Oops, this change is not passing all of my tests yet.

jpritikin:
Actually these changes are good. The problem I was having was caused by use of ftol instead of xtol on a constrained problem.

jpritikin:
How does this look?

jpritikin:
Hm, do I need to save/restore slsqpb_state if I pick a point from the line search that is not the last point?

stevengj:
@jpritikin, regarding the SAVE/RESTORE_STATE: if I remember correctly, the original Fortran code was using the equivalent of `static` C local variables for these things, and I wanted it to be re-entrant & thread-safe.   So, the state only needs saving inside the `slsqpb_`, and should be saved at every exit point from that function (unless you will never come back, e.g. you are exiting because you have converged or there was an error).

stevengj:
Could you squash these (via `git rebase --interactive` followed by a `git push -f`) into a single commit?  Also be sure to put `(for #30)` in the commit message so that #30 will be cross-referenced.

jpritikin:
> regarding the SAVE/RESTORE_STATE

This question was addressed with https://github.com/jpritikin/nlopt/commit/356bd27e08bbd20abb07534b6678257492ec8421

stevengj:
Should the `s[i__] = x[i__] - x0[i__];` loop be executed every time we `goto L260`, or only from the beginning of the `slsqpb_` function?  Maybe you should move it before the relevant `goto`.

jpritikin:
> Should the s[i__] = x[i__] - x0[i__]; loop be executed every time we goto L260, or only from the beginning of the slsqpb_ function? 

Sorry, I can't remember. Certainly the code doesn't hurt. It helps preserve an invariant condition that should be preserved.

jschueller:
hello @jpritikin, do you think this has a chance to land someday or do you think we can close it ?

jpritikin:
I personally have given up. The changes are part of OpenMx. I believe the version of the code in OpenMx is superior to what I started with when I copied it from here, but I haven't kept track of nlopt and I have also ceased to contribute to OpenMx.

jschueller:
thanks

PR #41, opened 2015-07-07
Added example of automatic tests

boris-il-forte:
- Added new_test target to generate test executable
- Updated CMakeLists.txt to add test as subfolder
- Added two test using executable new_test
- Now you can run test suite by just typing either "make test" or "ctest" in the build directory

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -173,3 +173,10 @@ INSTALL ( TARGETS nlopt
         )
 
 add_subdirectory (swig)
+
+# Tests
+add_subdirectory (test)
+enable_testing ()
+add_test (NAME test_lfbgs COMMAND new_test lfbgs)
+add_test (NAME test_mma COMMAND new_test mma)
+
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1,0 +1,3 @@
+add_executable (new_test newTest.cpp)
+target_link_libraries(new_test nlopt)
+ 
--- a/test/newTest.cpp
+++ b/test/newTest.cpp
@@ -1,0 +1,83 @@
+#include "nlopt.hpp"
+#include <iostream>
+#include <string>
+
+#include <cassert>
+
+double f(unsigned int n, const double* x, double* grad, void*)
+{
+    static int it = 0;
+
+    double a = x[0];
+    double b = x[1];
+    double c = x[2];
+
+    grad[0] = 2*a*b + 2*a*c + b*b + c*c + 1;
+    grad[1] = a*a + c*c + 2*b*a + 2*b*c + 1;
+    grad[2] = a*a + b*b + 2*c*a + 2*c*b + 1;
+
+    std::cout << it++ << std::endl;
+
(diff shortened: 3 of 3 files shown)


boris-il-forte:
This is only an example of adding automated test for nlopt, using cmake. The main issue is that now nlopt doesn't build automatically using cmake, autotools are needed. It would be nice to make first a pure cmake build: this can be possible simply adding the code generation in the CMakeLists.txt.

If you can point out the commands to trigger autogeneration or, even better, the makefiles used in the process, I can enable full build from cmake withouth the need of autogenerated files in the repository.

Also I can implement new tests or adapt the existing ones to the ctest framework to have a clean and automated test suite.

stevengj:
@boris-il-forte, the automake build is the primary one for development.  I would prefer that the tests be added there.

The `cmake` build is mainly only for Windows users.

stevengj:
Do these tests actually pass, due to #40?

stevengj:
There is also a built-in `testopt` program; the simplest thing would be to have a shell script that runs it automatically with various options.

boris-il-forte:
Personally, I use only linux and I only use cmake, whenever possible, as is easier and cleaner wrt autotools.
I don't know any automatic test suite framework that can be used with autotools, but they exist for sure.
However I'm not able to use autotools properly, so I cannot help with that.
Due to #40, actually one test pass (MMA) and one test fails (LBFGS). However pass/fail is not so well defined for my sample test: simply a pass occurs where there are no exceptions (and main return 0).

Of course you can define a well defined problem with a known optimum and check if the optimization algorithm works correctly by measuring some distance from the solution and returning 0 (pass) or -1 (fail) when the condition is not met. Or you can use assertion or exception as well.

stevengj:
@boris-il-forte, automake has built-in support for test programs (see the manual).  But you can also simply just write a test script that is executed by `make check`.

jschueller:
:+1:  for this, would you consider dropping autotools ?

boris-il-forte:
I was looking for Autotools test suite, and I see that is unnecessarily complex and unclean. So I'm not going to implement that in an autotools-like fashion. I can consider to write a shell script when I'll have time...

(Personally I think autotools are extremely unclean and unnecessarily complex...)

stevengj:
There's a fair amount of functionality in the `configure.ac` file; it would be hard to convince me that you've replicated all of this in `cmake`.

jschueller:
I see it's missing guile/octave/matlab bindings, and pkgconfig generation, got something else in mind ?

stevengj:
https://github.com/stevengj/nlopt/blob/master/m4/ax_c_threadlocal.m4, to pick the first example that pops out on a random scan.

stevengj:
I guess it's not completely crazy.  Autotools deals with a lot more ancient Unix variations than cmake, but these days that is no longer so relevant.

jschueller:
@stevengj, see https://github.com/jschueller/nlopt/commit/2d1122bfb752c61d198688200907802f810421bc

jschueller:
all these are implemented in #49

PR #42, opened 2015-07-08
Use octave_map

opoplawski:
This supports octave 4.0 and works at least in octave 3.8 if not earlier.

### DIFF
--- a/octave/nlopt_optimize-oct.cc
+++ b/octave/nlopt_optimize-oct.cc
@@ -29,7 +29,7 @@
 #include "nlopt.h"
 #include "nlopt_optimize_usage.h"
 
-static int struct_val_default(Octave_map &m, const std::string& k,
+static int struct_val_default(octave_map &m, const std::string& k,
 				 int dflt)
 {
   if (m.contains(k)) {
@@ -39,7 +39,7 @@ static int struct_val_default(Octave_map &m, const std::string& k,
   return dflt;
 }
 
-static double struct_val_default(Octave_map &m, const std::string& k,
+static double struct_val_default(octave_map &m, const std::string& k,
 				 double dflt)
 {
   if (m.contains(k)) {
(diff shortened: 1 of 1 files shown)


stevengj:
We should be able to continue to support earlier versions if you add:
```
#include <octave/version.h>
#if OCTAVE_MAJOR_VERSION < 3 || (OCTAVE_MAJOR_VERSION == 3 && OCTAVE_MINOR_VERSION < 8)
#  define octave_map Octave_map
#endif
```

opoplawski:
Actually, octave_map is present in octave 3.2, which is what is in EPEL5. So I don't think there is any need.

stevengj:
I'd prefer to just keep backward compatibility since there is almost no cost to doing so (just four lines).

opoplawski:
I'll let you do that then.  But the original change should be compatible with octave 3.2 without the conditional (as it has octave_map) which is 6 years old.

acapobianchi:
@stevengj
Hello, first of all thank you for NLopt!

I'm trying to compile Octave 4.0 plugin with the *mkoctfile* command, as explained at http://ab-initio.mit.edu/wiki/index.php/NLopt_on_Windows

I've generated the library *libnlopt-0.lib* and, in file *nlopt_optimize-oct.cc*, I've replaced all *Octave_map* with *octave_map*.

Unfortunately the command:

*mkoctfile -lD:\Installers\NLopt\nlopt-2.4.2-dll64\libnlopt-0.lib --output nlopt_optimize.oct D:\Installers\NLopt\nlopt-2.4.2-dll64\nlopt_optimize-oct.cc*

fails, returning:

c:/octave/octave-4.0.0/bin/../lib/gcc/i686-w64-mingw32/4.9.2/../../../../i686-w64-mingw32/bin/ld.exe: cannot find -lD:\Installers\NLopt\nlopt-2.4.2-dll64\libnlopt-0.lib
collect2.exe: error: ld returned 1 exit status
warning: mkoctfile exited with failure status
warning: called from
    mkoctfile at line 171 column 5

I've checked that the .lib file exist in the specified folder.

Are there more detailed instructions for Octave plugin compilation?

Thank you very much!

PR #46, opened 2015-08-31
Add Travis script

jschueller:
Add a travis continuous integration script, it will allow next commits & pull requests to test compilation.
It just needs the service "Travis CI" to be enabled in the repository settings.


.travis.yml
```diff
@@ -0,0 +1,19 @@
...
+    - libguile-dev
+    - octave-headers
+    - cmake
+
+script:
+  - ./autogen.sh --no-configure
+  - ./configure --prefix=$HOME/.local --enable-shared --enable-maintainer-mode --enable-cxx
+  - make -j2
+  - make install
+  - python ./test/test_std.py
```
stevengj: Don't you need to set up the `PYTHONPATH` to look in `$HOME/.local` for this to work?

.travis.yml
```diff
@@ -0,0 +1,19 @@
...
+  - cmake -DCMAKE_INSTALL_PREFIX=~/.local -DBUILD_SHARED_LIBS=ON .
```
stevengj: Normally you should `cmake` in a subdirectory:
```
mkdir build
cd build
cmake -DCMAKE_INSTALL_PREFIX=~/.local -DBUILD_SHARED_LIBS=ON ..
make -j2
```

.travis.yml
```diff
@@ -0,0 +1,19 @@
...
+    packages:
+    - swig
+    - python-dev
+    - libguile-dev
+    - octave-headers
+    - cmake
+
+script:
+  - ./autogen.sh --no-configure
+  - ./configure --prefix=$HOME/.local --enable-shared --enable-maintainer-mode --enable-cxx
```
stevengj: Since the makefiles support building in a subdirectory, that would be good to test:
```
mkdir build
cd build
../configure --prefix=$HOME/.local --enable-shared --enable-maintainer-mode --enable-cxx
make -j2
make install
```
and so on.

stevengj:
(I've enabled Travis for when this is ready to merge.)

stevengj:
Would be good to also include a Travis badge at the top of the `README.md` file

.travis.yml
jschueller: it's the default user python search path, used when you do "python setup.py install --user"

jschueller:
ok!

jschueller:
ok!

.travis.yml
```diff
@@ -0,0 +1,24 @@
...
+
+script:
+  - ./autogen.sh --no-configure
+  - mkdir build && pushd build
+  - ../configure --prefix=$HOME/.local --enable-shared --enable-maintainer-mode --enable-cxx
+  - make -j2
+  - make install
+  - python ../test/test_std.py
+  - rm -rf * ~/.local
+  - cmake -DCMAKE_INSTALL_PREFIX=~/.local -DBUILD_SHARED_LIBS=ON ..
```
stevengj: Would be good to use a separate build directory for `cmake`, to make sure it isn't accidentally using any `.o` files compiled by the first `make`.

stevengj:
Looking good, but I would prefer that the commits were squashed before merging.

.travis.yml
jschueller: the content of the build dir is removed : rm -rf * ~/.local

jschueller:
done

PR #49, opened 2015-09-04
Switch to cmake

jschueller:
This removes autotools and adds missing features to cmake.
It superseeds PRs #41, #67.
Some tests are run for win32/msvc 2015, linux/gcc and osx/clang.


jschueller:
Done. It superseeds PRs #41, #67.
Tests are run for win32/msvc 2015, linux/gcc and osx/clang.

CMakeLists.txt
```diff
@@ -14,112 +12,177 @@ PROJECT(NLOPT)
...
+check_include_file (sys/types.h HAVE_SYS_TYPES_H)
+check_include_file (sys/stat.h HAVE_SYS_STAT_H)
+check_include_file (sys/time.h HAVE_SYS_TIME_H)
+if (HAVE_TIME_H AND HAVE_SYS_TIME_H)
+  set (TIME_WITH_SYS_TIME TRUE)
+endif ()
+check_function_exists (getpid HAVE_GETPID)
+check_function_exists (syscall HAVE_GETTID_SYSCALL)
+check_function_exists (isinf HAVE_ISINF)
+check_function_exists (isnan HAVE_ISNAN)
```
stevengj: On some systems, I vaguely recall that isinf/isnan are macros rather than functions.  Does `check_function_exists` handle this case?

stevengj: Oh, I see that you had to handle this separately with `fpclassify` below.

CMakeLists.txt
```diff
@@ -140,41 +203,147 @@ SET ( NLOPT_SOURCES
...
+set (CPACK_BINARY_TGZ            "ON"         )
+set (CPACK_BINARY_TZ             "OFF"        )
+set (CPACK_SOURCE_IGNORE_FILES ".git;/build;.*~;${CPACK_SOURCE_IGNORE_FILES}")
+
+set (CPACK_SOURCE_PACKAGE_FILE_NAME ${CPACK_PACKAGE_NAME}-${NLOPT_VERSION_STRING})
+
+include (CPack)
+
+# write .def file
+if (MINGW AND BUILD_SHARED_LIBS)
```
stevengj: This seems like it may not be necessary with cmake?  The reason I was building the `.def` file is that the previous build script created the `.dll` when cross-compiling but didn't create the corresponding `.lib` import library, and people needed the `.def` file to create the `.lib` as described http://ab-initio.mit.edu/wiki/index.php/NLopt_on_Windows

I'm hoping that cmake is able to build the import library itself?

cobyla/Makefile.am
```diff
@@ -1,6 +0,0 @@
-AM_CPPFLAGS = -I$(top_srcdir)/util -I$(top_srcdir)/api
-
-noinst_LTLIBRARIES = libcobyla.la
-libcobyla_la_SOURCES = cobyla.c cobyla.h
```
stevengj: How does cmake know what files to compile in each directory?

jschueller: Each source file is listed in the main CMakeLists.txt.

CMakeLists.txt
jschueller: Yes CMake knows how to create the import library, so I guess this is useless.

CMakeLists.txt
jschueller: I suppose it doesn't make a difference if it's a macro or a function.

stevengj:
Okay, merged....

PR #63, opened 2016-04-12
add missing linking for C++

thrasibule:
When building the C++ library, some symbols were missing because libtool didn't use the right tag.
Also the python extension wasn't linked to the python libraries.


stevengj:
Looks good, thanks.

PR #67, opened 2016-06-03
Generate missing nlopt.hpp and nlopt.f (CMake), use GNUInstallDirs (CMake), fix for MSVC 2015

rickertm:
Create missing nlopt.hpp and nlopt.f when building with CMake
Use CMake's GNUInstallDIrs (e.g, ${CMAKE_INSTALL_LIBDIR} instead of lib) depending on platform
Fix for MSVC 2015 compiler

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,6 +87,56 @@ MESSAGE(STATUS "NLOPT: Version number ${NLOPT_MAJOR_VERSION}.${NLOPT_MINOR_VERSI
 
 CONFIGURE_FILE(${CMAKE_CURRENT_SOURCE_DIR}/config.cmake.h.in ${CMAKE_CURRENT_BINARY_DIR}/config.h IMMEDIATE )
 
+#==============================================================================
+# CREATE nlopt.hpp and nlopt.f
+#==============================================================================
+
+FILE(READ ${CMAKE_CURRENT_SOURCE_DIR}/api/nlopt.h NLOPT_H_FILE)
+
+#------------------------------------------------------------------------------
+# CREATE nlopt.hpp
+#------------------------------------------------------------------------------
+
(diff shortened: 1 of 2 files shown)


stevengj:
I'm not sure I want to maintain two completely redundant compilation methods.

My feeling up to now has been to require autotools for developers (people working with the git version), and only support CMake for the shipped tarball (which already includes the generated code).

rickertm:
I understand the overhead in maintaining two systems. For Windows builds with MSVC, autotools are a big issue and the tarball didn't include the CMake files for building an external project in CMake.

The CMake code reads the algorithm list from api/nlopt.h and updates api/nlopt-in.h and api/nlopt.f in the build directory.

stevengj:
We could just include the CMakeLists files in the tarball.

rickertm:
I'm currently trying to provide better (automatic) build support of our third-party dependencies for our users, especially for Windows with MSVC. Adding the CMake files to the tarball would address this. I certainly don't want to change your development process.

stevengj:
This was done in 260b6f700c80b12d2e47bc48233a6a4553866821 ... I really need to just put a new release together.

jschueller:
Now that #49 allows to build language bindings the cmake build system is on par with the level of features autotools supports (with the capabilty to build with non-unix system and other non-make systems), maybe dropping autotools could remove the hassle of supporting both systems.

stevengj:
I'm not averse to switching completely over to CMake, I guess, since CMake is pretty full-featured these days.   Can @jschueller and @rickertm (and maybe @boris-il-forte, from #41), join forces and make a single PR that combines your work as needed, removes the autotools scripts, and ideally replaces the `BUILD-mingw` scripts with CMake toolchain files for cross-compiling?

jschueller:
@stevengj I'm on it in #49, could you enable Appveyor continuous integration so as I could test on msvc 2015 ? I already added a config file.

stevengj:
Okay, just turned on AppVeyor.

stevengj:
Superseded by #49, I think?

PR #72, opened 2016-07-12
Readme

jschueller:


README.md
```diff
@@ -10,11 +10,11 @@ home page: http://ab-initio.mit.edu/nlopt
 
 It is compiled and installed with CMake build system:
 
-    cmake .
+    cmake -DCMAKE_INSTALL_PREFIX=~/.local .
```
stevengj: Why should we recommend this directory?

jschueller: @stevengj it's user-writeable (/usr/local is not), also python looks for user modules there so you don't have to set the PYTHONPATH env variable to use nlopt python bindings.

stevengj: Interesting, I didn't know that about Python.  Unfortunately, it is still not in the standard compiler search path so you will need `-I` and `-L` flags to compile C code with nlopt; maybe a note to that effect would be useful.

(`/usr/local` is the GNU standard default prefix for `make install` scripts, because it is the standard system-wide install location for things that are compiled from source.  Of course, it assumes you can do `sudo make install`)

jschueller: @stevengj, ok I wrote sudo make instal instead

stevengj:
Looks good, go ahead and merge when the tests are green.

stevengj:
In general, it would be better to use a separate PR to update the tests; don't mix it with the PR to update the documentation.

stevengj:
Each PR should ideally be a logically distinct change; that makes it easier to review.

stevengj:
Thanks!

jschueller:
@steveng of course, I was playing around with travis.

PR #73, opened 2016-07-13
Update tests

jschueller:


stevengj:
LGTM.

CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ option (WITH_CXX "enable cxx routines" OFF)
 option (BUILD_SHARED_LIBS "Build NLOPT as a shared library" ON)
 option (BUILD_PYTHON "build python bindings" ON)
 option (BUILD_OCTAVE "build octave bindings" ON)
+option (BUILD_MATLAB "build matlab bindings" OFF)
```
stevengj: Maybe we should build the Matlab plugin by default if `mex` is found?  That's what the old `configure` script did if I remember correctly.

PR #75, opened 2016-07-25
add cmake flag for testing, add common definitions file

pkarasev3:
- added some cmake helper functionality . 
- added BUILD_TESTS flag to configure from gui .
- added ctest hint so one can see the nicely done matrix of tests available.

### DIFF
--- a/cmake/common.cmake
+++ b/cmake/common.cmake
@@ -1,0 +1,35 @@
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/lib")
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/lib")
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/bin")
+    
+string(ASCII 27 Esc)
+set(Green    "${Esc}[92m"     CACHE STRING "green color"    FORCE)
+set(Blue     "${Esc}[1;94m"   CACHE STRING "Blue color"     FORCE)
+set(BoldRed  "${Esc}[1;5;91m" CACHE STRING "bold-red color" FORCE)
+set(NoColor  "${Esc}[0m"      CACHE STRING "reset color"    FORCE)
+
+if(NOT WIN32)
+
+    execute_process(COMMAND   ${CMAKE_C_COMPILER}  -dumpversion   OUTPUT_VARIABLE   GCC_VERSION)
+    if(GCC_VERSION VERSION_GREATER 4.9 OR GCC_VERSION VERSION_EQUAL 4.9)
+        set(STDLIBC  "-std=c++1y")
+    elseif(GCC_VERSION VERSION_EQUAL 4.8)
+        set(STDLIBC  "-std=c++11")
+    else()
+        set(STDLIBC  " ")
+    endif()
+    set(STDLIBCXX_FLAG  ${STDLIBC}  CACHE STRING "stdlibc++ flag")
+    message(STATUS "${Green}GCC_VERSION= ${GCC_VERSION} ")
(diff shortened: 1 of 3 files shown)


cmake/common.cmake
```diff
@@ -0,0 +1,35 @@
...
+set(Green    "${Esc}[92m"     CACHE STRING "green color"    FORCE)
+set(Blue     "${Esc}[1;94m"   CACHE STRING "Blue color"     FORCE)
+set(BoldRed  "${Esc}[1;5;91m" CACHE STRING "bold-red color" FORCE)
+set(NoColor  "${Esc}[0m"      CACHE STRING "reset color"    FORCE)
+
+if(NOT WIN32)
+
+    execute_process(COMMAND   ${CMAKE_C_COMPILER}  -dumpversion   OUTPUT_VARIABLE   GCC_VERSION)
+    if(GCC_VERSION VERSION_GREATER 4.9 OR GCC_VERSION VERSION_EQUAL 4.9)
+        set(STDLIBC  "-std=c++1y")
```
stevengj: Why do we need this?  C++ 2003 is probably fine.... we don't use any recent C++ features, so there should be no need to specify `-std=c++XX`.

cmake/common.cmake
```diff
@@ -0,0 +1,35 @@
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/lib")
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/lib")
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/bin")
+    
+string(ASCII 27 Esc)
+set(Green    "${Esc}[92m"     CACHE STRING "green color"    FORCE)
+set(Blue     "${Esc}[1;94m"   CACHE STRING "Blue color"     FORCE)
```
stevengj: Do these terminal codes work on Windows?

cmake/common.cmake
pkarasev3: "we don't use ..."  presumably that could change right? :-)  

 It's nice to be able to use bounds-checking structures with simple syntax. 
 For example: 
       `**std::vector<double> coefficients = {1.0, 2.0,  rand()*1.0 };** ` 
  looks a lot simpler than however you'd do it in C++03 (for-loop?), and safer than a raw C-style array. 

Also, on Windows, visual studio 2012 by default will enable C++11, while visual studio 2013 enables C++14. These gcc flags are to align the default compiler behavior across platforms.

cmake/common.cmake
```diff
@@ -0,0 +1,35 @@
...
+set(BoldRed  "${Esc}[1;5;91m" CACHE STRING "bold-red color" FORCE)
+set(NoColor  "${Esc}[0m"      CACHE STRING "reset color"    FORCE)
+
+if(NOT WIN32)
+
+    execute_process(COMMAND   ${CMAKE_C_COMPILER}  -dumpversion   OUTPUT_VARIABLE   GCC_VERSION)
```
jschueller: This is not really portable "-dumpversion" is only available with GCC (maybe clang).

test/CMakeLists.txt
```diff
@@ -1,7 +1,13 @@
 add_custom_target (tests)
 
+if(BUILD_TESTS)
+    set(TESTING_FLAG  "")
+else()
+    set(TESTING_FLAG EXCLUDE_FROM_ALL)
+endif()
+
```
jschueller: wouldn't be easier to exlude the compiled tests in all cases and provide a custom target ?

cmake/common.cmake
stevengj: We use C++ so little in NLopt that the benefits of using features from the latest C++ standard don't seem to outweigh the cost of dropping support for usage by old programs that require earlier versions of the standard.

PR #76, opened 2016-08-07
Use CMake module GNUInstallDirs to support platform-specific installation directories

rickertm:
The switch to CMake looks great, thank you very much. The only thing that got lost is the use of the GNUInstallDirs module from CMake to support platform-specific installation directories. This is especially noticeable for LIBDIR (possible values include 'lib', 'lib64', 'lib/x86_64-linux-gnu' depending on the target platform). It requires a minimum CMake version of 2.8.5 however.


jschueller:
+1

stevengj:
Thanks, LGTM.

PR #71, opened 2016-07-07
improve the MATLAB bindings

amroamroamro:
Several improvements to the mex-file.
Commit messages give a summary of the changes.

PR #83, opened 2016-09-12
patch for CMake & MATLAB

amroamroamro:
This PR contains updates to the cmake build scripts, in addition to improvements to the MATLAB MEX-file (this part is mostly the same as #71 I submitted a while back before the transition from autotools to cmake).

Changes are described in each commit message, but as a quick summary:
- first few commits introduce several small fixes
- next bulk of commits update the cmake scripts
- final part improves the matlab module


amroamroamro:
had to make quick fixes, but it's all green now :)

README.md
```diff
@@ -23,10 +23,10 @@ To build the latest development sources:
 
 (To build from git, you will need SWIG.)
 
-Once it is installed, `#include <nlopt.h>` in your C/C++ programs and
+Once it is installed, `#include "nlopt.h"` in your C/C++ programs and
```
jschueller: no, keep the brackets

CMakeLists.txt
```diff
@@ -483,14 +485,12 @@ install (
 #==============================================================================
 
 # create and install pkgconfig file
-if (UNIX OR MINGW)
-  configure_file (nlopt.pc.in nlopt.pc @ONLY)
+configure_file (nlopt.pc.in nlopt.pc @ONLY)
```
jschueller: so nlopt.pc gets configured/installed even on msvc ?

README.md
amroamroamro: I suppose there is little difference, but it seems weird to use `<...>` for non-system library headers.

http://stackoverflow.com/questions/21593/what-is-the-difference-between-include-filename-and-include-filename

CMakeLists.txt
amroamroamro: you can imagine compiling the library using MSVC, and later using it from MinGW with pkg-config.
Hence I installed the .pc file, same thing with the man pages..

CMakeLists.txt
```diff
@@ -107,6 +107,17 @@ if (CMAKE_C_COMPILER_ID STREQUAL "GNU" OR CMAKE_C_COMPILER_ID STREQUAL "Clang")
...
+# in order to avoid dependencies on libgcc and libstdc++ shared libraries
+# (this produces self-contained but larger binaries)
+# (NOTE: pay attention if C++ exceptions are propagated across libraries)
+if (CMAKE_CROSSCOMPILING AND CMAKE_COMPILER_IS_GNUCC AND BUILD_SHARED_LIBS)
+  set (CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -static")
+  set (CMAKE_STATIC_LINKER_FLAGS "${CMAKE_STATIC_LINKER_FLAGS} -static")
+  set (CMAKE_MODULE_LINKER_FLAGS "${CMAKE_MODULE_LINKER_FLAGS} -static")
+  set (CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -static")
+endif ()
+
```
jschueller: That should be left to the developer to specify this or not, for example I'm used to use shared libs.

README.md
stevengj: nlopt.h is installed into system include paths.   You normally use `"...."` only for include files within your own project, not for external libraries like nlopt.

CMakeLists.txt
```diff
@@ -1,330 +1,712 @@
...
-
-configure_file (cmake/NLoptConfig.cmake.in NLoptConfig.cmake @ONLY)
-configure_file (cmake/NLoptConfigVersion.cmake.in NLoptConfigVersion.cmake @ONLY)
-install (FILES
-          ${CMAKE_CURRENT_BINARY_DIR}/NLoptConfig.cmake
-          ${CMAKE_CURRENT_BINARY_DIR}/NLoptConfigVersion.cmake
-         DESTINATION ${RELATIVE_INSTALL_CMAKE_DIR}
-         COMPONENT Development)
-
+#==============================================================================
```
stevengj: What's going on here with the diff?  Did you change line endings or something?  Why does it show every line has changed?

CMakeLists.txt
```diff
@@ -1,330 +1,712 @@
...
+# AUTHOR: Steven G. Johnson
+#
+# This CMakeLists.txt file was created to compile NLOPT with the CMAKE utility.
+# Benoit Scherrer, 2010 CRL, Harvard Medical School
+# Copyright (c) 2008-2009 Children's Hospital Boston
+#==============================================================================
+
+cmake_minimum_required (VERSION 2.8.9)
+
+# default install directory locally inside build-tree
```
stevengj: I don't see the point of this.  If you're going to do `make install` at all, normally you would want it to be in a more globally visible directory, not inside the build tree — that's the whole point of `make install`.    The GNU standard default is `/usr/local`

CMakeLists.txt
```diff
@@ -38,11 +38,12 @@ list (APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
 # provide user options
 option (BUILD_SHARED_LIBS "Build NLopt as a shared library" ON)
 option (WITH_CXX          "Enable C++-based routines (StoGO)" OFF)
-option (USE_SWIG          "Use SWIG to generate bindings" ON)
+option (WITH_SWIG          "Use SWIG to generate bindings" ON)
 option (BUILD_PYTHON      "Build Python bindings (using SWIG)" ON)
 option (BUILD_GUILE       "Build Guile bindings (using SWIG)" ON)
 option (BUILD_OCTAVE      "Build Octave bindings" ON)
 option (BUILD_MATLAB      "Build MATLAB bindings" ON)
+option (BUILD_TESTS       "Build Tests" ON)
```
jschueller: this is useless, as by default the test executable is not built (excluded from all target)

CMakeLists.txt
```diff
@@ -1,330 +1,712 @@
...
+
+# append path to custom cmake modules FindXXX
+list (APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
+
+# provide user options
+option (BUILD_SHARED_LIBS       "Build NLopt as a shared library" ON)
+option (WITH_CXX                "Enable C++-based routines (StoGO)" OFF)
+option (WITH_SWIG               "Use SWIG to generate bindings" ON)
+option (BUILD_CPP               "Build C++ interace (nlopt.hpp)" ON)
+option (BUILD_F77               "Build Fortran iterface" ON)
```
stevengj: iterface -> interface

CMakeLists.txt
```diff
@@ -1,330 +1,712 @@
...
+option (BUILD_PYTHON            "Build Python bindings (using SWIG)" ON)
+option (BUILD_GUILE             "Build Guile bindings (using SWIG)" ON)
+option (BUILD_OCTAVE            "Build Octave bindings" ON)
+option (BUILD_MATLAB            "Build MATLAB bindings" ON)
+option (BUILD_TESTS             "Build tests" ON)
+option (BUILD_TESTS_UTIL        "Build util tests" OFF)
+option (BUILD_TESTS_ALG         "Build algorithms tests" OFF)
```
stevengj: Why should this default to OFF?

In general, I don't think it's worthwhile to have a lot of fine-grained build options like this.  The only things that should be build options are the things that depend on external libraries, like Python or Guile bindings.

So, for example, we can unconditionally build `nlopt.hpp` because that doesn't depend on any external libraries, or even on the presence of a C++ compiler.  Similarly for the Fortran interface.  Similarly from the tests.

CMakeLists.txt
```diff
@@ -1,330 +1,712 @@
...
+      set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${_FLG}")
+    endforeach ()
+    unset (_FLG)
+  else ()
+    # turn off deprecated warnings
+    set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-deprecated-declarations")
+    set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-deprecated-declarations")
+  endif ()
+endif ()
+
```
jschueller: It's simpler to let each developer override its own flags. Wshadow is buggy on gcc anyway.

stevengj:
In general, it is quite hard to review a PR that includes so many unrelated changes.   Each PR should add one feature or fix one problem.

CMakeLists.txt
amroamroamro: I see, I will remove this change.

My thought at the time was that the default installation directory on Windows, being "Program Files" which is system-protected, is not developer friendly and forces you to run the whole cmd.exe shell or Visual Studio instance as administrator (no easy equivalent like running `sudo make install`)...

README.md
amroamroamro: I'll undo my change.

CMakeLists.txt
amroamroamro: I guess we could add an `option()` to control this ON or OFF when cross-compiling..

But just to be clear, the `-static` link flag does not mess with `BUILD_SHARED_LIBS` setting, a shared library (DLL) will still be generated if it was requested (i.e `-static` and `-shared` GCC flags can be used together). It is only telling the linker not to depend on MinGW DLLs same as using the flags `-static-libgcc` and `-static-libstdc++`:

> https://gcc.gnu.org/onlinedocs/gcc/Link-Options.html#index-static-1164

To illustrate, without using `-static` flag we get this:

``` sh
# `objdump` is used here as a poor substitute of `ldd` for windows binaries

# WITH_CXX=OFF BUILD_SHARED_LIBS=ON
$ x86_64-w64-mingw32-objdump -p libnlopt.dll | grep 'DLL Name'
    DLL Name: KERNEL32.dll
    DLL Name: msvcrt.dll
    DLL Name: libgcc_s_sjlj-1.dll

# WITH_CXX=ON BUILD_SHARED_LIBS=ON
$ x86_64-w64-mingw32-objdump -p libnlopt_cxx.dll | grep 'DLL Name'
    DLL Name: KERNEL32.dll
    DLL Name: msvcrt.dll
    DLL Name: libgcc_s_sjlj-1.dll
    DLL Name: libstdc++-6.dll
```

Using `-static` we get this:

``` sh
# WITH_CXX=OFF BUILD_SHARED_LIBS=ON
$ x86_64-w64-mingw32-objdump -p libnlopt.dll | grep 'DLL Name'
    DLL Name: KERNEL32.dll
    DLL Name: msvcrt.dll

# WITH_CXX=ON BUILD_SHARED_LIBS=ON
$ x86_64-w64-mingw32-objdump -p libnlopt_cxx.dll | grep 'DLL Name'
    DLL Name: KERNEL32.dll
    DLL Name: msvcrt.dll
```

So without `-static` those extra libgcc and libstdc++ dependencies will have to be copied as well when deploying cross-compiled binaries.

Of course the downside is that the binaries are larger.

CMakeLists.txt
amroamroamro: Right, I did change CRLF to LF: https://github.com/stevengj/nlopt/pull/83/commits/160a105f6940f359725696ad335337d78482ccd8
I should have done it in a separate commit, sorry about that!

One [tip](https://github.com/blog/967-github-secrets) I recently learned is to add `?w=1` to the Github URL to ignore whitespace changes in diffs: https://github.com/stevengj/nlopt/pull/83/commits/160a105f6940f359725696ad335337d78482ccd8?diff=split&w=1

CMakeLists.txt
amroamroamro: noted, I'll remove the extra options (BUILD_TESTS* and BUILD_CPP).

I would argue to keep the Fortran one `BUILD_F77`, since it does involve exporting new symbols in the library, and there are valid cases where one would prefer a smaller executable/symbol-table if Fortran-compatible API is not being used.

CMakeLists.txt
amroamroamro: thanks, I'll remove it and the other `BUILD_TESTS_*` ones.

CMakeLists.txt
amroamroamro: I actually got these from the old `configure.ac` file, the so called "debug/maintainer mode": https://github.com/stevengj/nlopt/blob/f8f167650d364334285d77a283c8bdd902a97de2/configure.ac#L397

I'll remove this block if no longer needed?

CMakeLists.txt
amroamroamro: thanks

amroamroamro:
@stevengj apologies for making it hard to review. Shall we continue with this PR or should I attempt to break it into smaller ones?

Originally I was going to create two PRs one for cmake and one for matlab, but since there was little overlap between them, I decided to push it as one (the matlab changes are simply rebased from the old PR with minor edits)..

jschueller:
please do break it

amroamroamro:
@jschueller Done, first one is up (#84) out of 5 or 6 patches total...

PR #84, opened 2016-09-14
Patch 1

amroamroamro:
patch includes several small fixes:
- update readme file
- fix a few compiler warnings in C files
- create tests that were previously commented inside `util` files
- move definition of functions used in stop-criteria to `util/stop.c`
- expand `isinf`/`isfinite` functions


stevengj:
LGTM.

PR #85, opened 2016-09-14
Patch 2: tests

amroamroamro:
Few fixes to test files:
- updated and simplified `lorentzfit`
- put `getopt` implementation in separate files
- fixes and updates to `testopt` + renamed as .c file

I also included the following edits to make sure it passes CI checks:
- explicitly link nlopt against libm
- fixed octave include dirs needed for octave 4.x


stevengj:
Please don't change the indenting.  That just creates unnecessary noise in the diff.

When making a PR, you should always match the existing indenting in file.

amroamroamro:
I get that but the changes were intentional, in order to fix inconsistent style of indentation. This includes:
- avoid mixing both spaces and tabs for indenting
- use a consistent number of spaces indented (either 2 or 4-spaces according to the existing file)
- trim trailing whitespaces
- split long lines

Although these are minor edits, I find that it makes the code more readable, don't you think?

amroamroamro:
By the way, I already have the rest of the changes split and ready to push, I'm just submitting PRs one by one so that I can rebase each time on top of the new master after each merge, that way the commit history stays clean and linear.

stevengj:
It makes sense to make the indentation more consistent, but not in a PR that makes other changes, because then it makes it hard to find the non-whitespace changes.   One PR that *only* makes indentation fixes is fine.

You should really have one PR per conceptually distinct issue.  Even if that means 15 PRs.

test/testopt.cpp
```diff
@@ -47,7 +47,7 @@ extern "C" int feenableexcept (int EXCEPTS);
 #include "testfuncs.h"
 
 static nlopt_algorithm algorithm = NLOPT_GN_DIRECT_L;
-static double ftol_rel = 0, ftol_abs = 0, xtol_rel = 0, xtol_abs = 0, minf_max_delta = -HUGE_VAL;
```
stevengj: I've just enabled appveyor on the repo.

jschueller: @stevengj could you pass me the appveyor url of nlopt ?

jschueller: oh, right, that was done some time ago.

jschueller:
mostly fixed in #136, lm was already added

PR #90, opened 2016-10-25
Fix OSX build on new image

jschueller:


.travis.yml
```diff
@@ -35,8 +35,8 @@ matrix:
         - brew tap homebrew/python
         - brew tap homebrew/science
         - brew update
-        - brew install swig numpy octave guile
+        - brew install swig octave guile || echo "nope"
       script:
         - mkdir build && pushd build
-        - cmake -DCMAKE_INSTALL_PREFIX=~/.local  -DWITH_CXX=ON ..
+        - cmake -DCMAKE_INSTALL_PREFIX=~/.local -DCMAKE_CXX_FLAGS="-std=c++11" -DWITH_CXX=ON -DPYTHON_EXECUTABLE=/usr/bin/python ..
```
stevengj: Shouldn't `-std=c++11` be automatically added by `cmake` if it is needed?  I don't think we want to force the user to specify this manually.

PR #93, opened 2016-10-27
Detect libcpp

jschueller:


CMakeLists.txt
```diff
@@ -132,6 +134,16 @@ if (WITH_THREADLOCAL AND NOT DEFINED HAVE_THREAD_LOCAL_STORAGE)
   endforeach()
 endif ()
 
+if (WITH_CXX OR BUILD_PYTHON OR BUILD_GUILE OR BUILD_OCTAVE)
+  check_cxx_symbol_exists (_LIBCPP_VERSION string SYSTEM_HAS_LIBCPP)
```
stevengj: Why do we need to check this?

stevengj: i.e. why not always use `-std=c++11` if we are compiling with C++ and that flag is supported?

jschueller: - c++ support is only enabled if requested or bindings enabled
- NLopt itself doesn't use c++11

jschueller: There's no problem with nlopt at all, I followed your remark about my travis script that has to set the -std=c++11 flag to compile on osx.

PR #103, opened 2016-11-04
Support cmake && make && ctest

### LINKED ISSUE #98: Running tests with CTest is broken
If I check out NLopt, build it, and try to run the tests with `ctest`, it fails horribly because `testopt` was not built.

The way this works is a non-standard workflow and prevents running the tests e.g. when NLopt is built as a CMake external project.

Please, either just build `testopt` by default (i.e. remove `EXCLUDE_FROM_ALL`), or else add an option (e.g. `NLOPT_ENABLE_TESTS`) that controls whether `testopt` is built by default *and* whether any `add_test` are invoked.

jschueller:
Fixes #98

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,13 +22,13 @@ matrix:
       script:
         - mkdir build && pushd build
         - cmake -DCMAKE_INSTALL_PREFIX=~/.local -DWITH_CXX=ON ..
-        - make install -j2 && make tests -j2 && ctest -j2 --output-on-failure
+        - make install -j2 && ctest -j2 --output-on-failure
         - rm -rf * ~/.local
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -22,13 +22,13 @@ matrix:
       script:
         - mkdir build && pushd build
         - cmake -DCMAKE_INSTALL_PREFIX=~/.local -DWITH_CXX=ON ..
-        - make install -j2 && make tests -j2 && ctest -j2 --output-on-failure
+        - make install -j2 && ctest -j2 --output-on-failure
```
stevengj: I'm confused, I thought that the point of this patch is that the tests for `ctest` are already built?  Why is `make tests` needed?

(Note also that you can just do `make -j2 install tests` to build multiple targets; this is better because that way it has more potential parallelism.)

jschueller: it's removed, see it's in red

stevengj: Oh, sorry, I read the patch backwards.

PR #108, opened 2016-11-22
Prefer target_include_directories in CMake build script

jdumas:
This PR facilitates the inclusion of `nlopt` into other CMake projects.

- The "private" include directories are defined in a per-target basis via `target_include_directories` instead of `include_directories`. This has the advantage that it doesn't "pollute" parent projects.
- Public include directories are set via the INTERFACE argument in `target_include_directories`. To keep it simple the original header `nlopt.h` is simply copied in the `${PROJECT_BINARY_DIR}/api` folder, which is used as the interface include directory for `nlopt`. Unfortunatelly, absolute paths cannot be given as interface include directories for installed targets, hence the need for the trick with `$<BUILD_INTERFACE:...>`. However this generator expression has been introduced in cmake 3.0, so the minimum required version is bumped to 3.0, but maybe you don't want that?
- Similarly, use `target_compile_definitions` instead of a global `add_definitions`.
- I couldn't make per-target include directories work with SWIG, so there is still a `include_directories` appearing in the `swig/CMakeLists.txt`. This is not ideal, but if you have any idea to improve this I'm all ears.

Now, building `nlopt` as part of other projects is as simple as
```cmake
add_subdirectory(ext/nlopt)
target_link_libraries(my_program nlopt)
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,8 @@ matrix:
     - os: linux
       addons:
         apt:
+          sources:
+          - george-edison55-precise-backports
           packages:
           - swig
           - python-dev
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ matrix:
           - guile-2.0-dev
           - octave3.2-headers
           - cmake
+          - cmake-data
           - g++-mingw-w64-i686
           - gcc-mingw-w64-i686
           - binutils-mingw-w64-i686
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,18 +1,18 @@
 #==============================================================================
 # NLOPT CMake file
-# 
-# NLopt is a free/open-source library for nonlinear optimization, providing 
-# a common interface for a number of different free optimization routines 
-# available online as well as original implementations of various other 
+#
+# NLopt is a free/open-source library for nonlinear optimization, providing
+# a common interface for a number of different free optimization routines
+# available online as well as original implementations of various other
 # algorithms
-# WEBSITE: http://ab-initio.mit.edu/wiki/index.php/NLopt 
+# WEBSITE: http://ab-initio.mit.edu/wiki/index.php/NLopt
 # AUTHOR: Steven G. Johnson
 #
 # This CMakeLists.txt file was created to compile NLOPT with the CMAKE utility.
 # Benoit Scherrer, 2010 CRL, Harvard Medical School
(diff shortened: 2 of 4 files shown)


stevengj:
Ideally, `swig` should not be required unless you are building from git master.  We should ship a .tar.gz file with the pre-generated SWIG wrappers.

jdumas:
Yes. In my case I am using `nlopt` as a git submodule in another project, so it is still useful to be able to `add_subdirectory()` with the git master branch. Git submodules are very convenient to handle dependencies to third-party libraries and keep them up-to-date.

jschueller:
please add this repo to the travis script so cmake gets updated to 3.0:
```
apt:
 sources:
  george-edison55-precise-backports
```

jdumas:
About CMake 3.0, quoting from my initial message:

> Unfortunatelly, absolute paths cannot be given as interface include directories for installed targets, hence the need for the trick with $<BUILD_INTERFACE:...>. However this generator expression has been introduced in cmake 3.0, so the minimum required version is bumped to 3.0, but maybe you don't want that?

jschueller:
maybe list cmake-data package too

jschueller:
for swig the target name is ${SWIG_MODULE_nlopt_REAL_NAME}

jdumas:
I have updated the CMakeList.txt, but I am not sure what you mean by the "cmake-data" package.

jschueller:
just try to add
```
- cmake-data
```
in the package section

jdumas:
I think you have also write access to my master branch, but I've updated it as you suggested.

jdumas:
Hmm, I have no idea why the build failed on Mac with Travis. Is that even related to this pull request?

jschueller:
It's related to guile support in swig: #116.

jdumas:
Done. I still can't get rid of the `include_directories (${NLOPT_PRIVATE_INCLUDE_DIRS})` in `swig/CMakeLists.txt`. If I comment it I still get a `Error: Unable to find 'nlopt.hpp'` despite this [`target_include_directories(...)`](https://github.com/jdumas/nlopt/blob/master/swig/CMakeLists.txt#L35) in the cmake script.

jdumas:
Let met know if anything else needs to be done before merging.

jdumas:
Hi guys,

Any chance of this PR being merged anytime soon? I don't see what's blocking this atm.

jdumas:
Thanks =)

Issue #127, opened 2017-07-14
Integer overflow in `nlopt_srand_time`


m-chaturvedi:
To reproduce, instrument the library with UBSan, and use the following code:
```
#include <bits/stdc++.h>
#include "nlopt.h"
using namespace std;

int main(){
nlopt_srand_time();
}
```
Results in an output similar to 
```
nlopt/api/general.c:122:50:runtime error: signed integer overflow: 27945 * 314159 cannot be represented in type 'int'
```
http://releases.llvm.org/3.9.0/tools/clang/docs/UndefinedBehaviorSanitizer.html

stevengj:
Closed by #128.

PR #128, opened 2017-07-14
Fixing Integer Overflow in nlopt_srand_time

m-chaturvedi:
The overflow happens due to operator precedence.
re https://github.com/stevengj/nlopt/issues/127

### DIFF
--- a/api/general.c
+++ b/api/general.c
@@ -119,7 +119,7 @@ void NLOPT_STDCALL nlopt_srand(unsigned long seed) {
 }
 
 void NLOPT_STDCALL nlopt_srand_time(void) {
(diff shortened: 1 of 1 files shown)


stevengj:
The overflow here is perfectly fine. The seed just needs to be a distinct integer, it doesn't matter what it is.

m-chaturvedi:
I see, thanks.
Perhaps a suppression would be alright, then?

jschueller:
I'd rather take the modulo 2**31

stevengj:
I'm fine with suppressing the warning, but I'd prefer to do it by casting to `unsigned long` rather than something MSVC-specific.

m-chaturvedi:
OK, then I'll revert to my previous commit.

Issue #36, opened 2015-04-27
Use of uninitialized variable in bobyqa altmov_()


gddl:
Under some circumstances the variable `ibdsav` is used uninitialized.
`altmov_()` in bobyqa.c
```C++
997    if (ibdsav < 0) {
998        xnew[-ibdsav] = sl[-ibdsav];
999    }
```

The following test code shows the situation that causes lots of NANs in `altmov_()` and leaves the `ibdsav` uninitialized. I have no idea, how the correct result for this input should look like - at least it shouldn't segfault. Tested on 64 Bit Linux architectures, gcc 4.8.2.

nlopt-test.cc
```C++
/* LN_BOBYQA test causing use of unitialized variable (64 Bit architecture only).
 */

#include <nlopt.hpp>
#include <vector>
#include <limits>
#include <xmmintrin.h>
#include <cstdio>

double callback(const std::vector<double>& x, std::vector<double>& grad, void* data)
{
    return 0.0010540095162719094;
}

int main()
{
    _mm_setcsr(_mm_getcsr() | (_MM_FLUSH_ZERO_ON));

    std::vector<double> lowerBounds(2, 0.0);
    std::vector<double> upperBounds(2, 0.0);
    lowerBounds[0] = -3.1415926535897931;
    lowerBounds[1] = -2.2250738585072014e-307;
    upperBounds[0] = 3.1415926535897931;
    upperBounds[1] = 2.4475812443579215e-307;

    nlopt::opt opt(nlopt::LN_BOBYQA, 2);

    opt.set_lower_bounds(lowerBounds);
    opt.set_upper_bounds(upperBounds);

    opt.set_min_objective(callback, 0);
    opt.set_xtol_rel(1.0e-8);

    double minf;
    std::vector<double> x(2, 0.0);
    x[0] = -0.59243378920273182;
    x[1] = -1.9140531208103923e-307;

    opt.optimize(x, minf);

    return 0;
}
```

stevengj:
Hmm, digging through the BOBYQA spaghetti code, it seems like `if (predsq > presav)` on line 971 should be `if (predsq >= presav)`.  Can you see if this fixes your problem?

stevengj:
[...]

PR #134, opened 2017-08-23
Fix all uninitialized variable warnings

### LINKED ISSUE #133: Bobyqa uninitialised variables in rare cases
Hello,

I know this is a follow-up to issue #36 but I do not know how to re-open it. I stumbled on the same problem (segmentation fault in altmov_ due to two variables being uninitialised) when using bobyqa together with the augmented lagrangian and a small tolerance as stopping criterion (10^-12). 

I ended up modifying the code of bobyqa a very little which made it work:

 - line 764 : int ksav = 1;
- line 768 : int ibdsav = 1;

This is only an ugly correction but I would be curious to know if someone has a better option to avoid this (I suspect all of this comes from the optimised parameters being a tiny little outside the bounds when entering bobyqa - from auglag - , which does not check for it)

Thanks

jschueller:
Every variables that yields a -Wmaybe-uninitialized gcc warning are set to 0

Fixes #133

### DIFF
--- a/bobyqa/bobyqa.c
+++ b/bobyqa/bobyqa.c
@@ -158,8 +158,8 @@ static nlopt_result rescue_(int *n, int *npt, const double *xl, const double *xu
     /* Local variables */
     double f;
     int i__, j, k, ih, jp, ip, iq, np, iw;
-    double xp, xq, den;
-    int ihp;
+    double xp = 0.0, xq, den;
+    int ihp = 0;
     double one;
     int ihq, jpn, kpt;
     double sum, diff, half, beta;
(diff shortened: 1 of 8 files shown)


ocommowi:
I tested (cf #36 comments) on my crashes and it does solve the problem on the few examples I have. Thanks again !

Looking at the code again I still have the feeling that variables used as indexes should be initialised to 1 rather than 0 for fortran codes converted to C (see in bobyqa.c, altmov_ function for example, all table pointers have a -- applied to them at the start of the function and thus index 0 should not be accessed). That's my humble opinion though, again I get lost easily with these indexing conventions problems

jschueller:
I dont know. I tried looking at the original code, zero are also used. What happens if you initialize it to 8000 for example ?

ocommowi:
Yes I agree too that the original code has an error. For me, the >0 and <0 branches of the code are there to handle possible negative indexes and make them positive again when searching in the tables, but 0 is not handled. That's also why I preferred to initialise at 1 but 0 is working for me, and as you said it will at least make it deterministic so it is good to do this initialization anyway.

jschueller:
that's really unclear, I propose we stick to 0 everywhere.

stevengj:
On the one hand, 0 is probably a good default when in doubt; this code was translated from Fortran (via f2c), and if I remember correctly there were ancient versions of Fortran in which variables were implicitly initialized to zero.  (The original author of this code, Powell, definitely harkens back to those days… it's very much Fortran-66 style spaghetti.)

On the other hand, you have to be careful about masking one bug with another bug by initializing to an arbitrary value... at least tools like valgrind can catch uses of uninitialized variables, but they can't catch incorrect initializations.

stevengj:
Looking at the code, it seems right to be to use `ibdsav = 0`, causing it to skip the final `xnew[±ibdsav] = ...` assignment.   The preceding loop is supposed to identify the dimension where `predsq` is biggest, and uses `ibdsav` to push `xnew` in that dimension to the boundaries of the trust region.  But if for some strange reason `predsq` is not biggest along any direction (maybe because it is always zero??? not sure why this would happen), then it seems right not to do any further modification of `xnew`.

jschueller:
ok thank you for peaking in

PR #137, opened 2017-08-24
Fix get|set lower|upper bounds methods warnings

### LINKED ISSUE #111: Warning with UndefinedBehaviorSanitizer
Hi,

Compiling nlopt with [UndefinedBehaviorSanitizer](http://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html), it produces a few warnings that would be easy to get rid of.

`api/general.c:122:50: runtime error: signed integer overflow: 7412 * 314159 cannot be represented in type 'int'`

- Caused by the product `my_gettid() * 314159` in `nlopt_srand_time()`.

`api/options.c:277:11: runtime error: null pointer passed as argument 1, which is declared to never be null`

- Which happens when calling `nlopt_set_lower_bounds` on a problem of size 0. It happens e.g. on the dual problem of `mma()` when there are 0 constraints.

jschueller:
fixes #111

### DIFF
--- a/api/options.c
+++ b/api/options.c
@@ -272,7 +272,8 @@ NLOPT_STDCALL nlopt_set_lower_bounds(nlopt_opt opt, const double *lb)
      nlopt_unset_errmsg(opt);
      if (opt && (opt->n == 0 || lb)) {
 	  unsigned int i;
-	  memcpy(opt->lb, lb, sizeof(double) * (opt->n));
+         if (opt->n > 0)
+             memcpy(opt->lb, lb, sizeof(double) * (opt->n));
 	  for (i = 0; i < opt->n; ++i)
               if (opt->lb[i] < opt->ub[i] && nlopt_istiny(opt->ub[i] - opt->lb[i]))
                   opt->lb[i] = opt->ub[i];
@@ -314,7 +315,8 @@ NLOPT_STDCALL nlopt_set_upper_bounds(nlopt_opt opt, const double *ub)
      nlopt_unset_errmsg(opt);
      if (opt && (opt->n == 0 || ub)) {
 	  unsigned int i;
-	  memcpy(opt->ub, ub, sizeof(double) * (opt->n));
+         if (opt->n > 0)
+             memcpy(opt->ub, ub, sizeof(double) * (opt->n));
 	  for (i = 0; i < opt->n; ++i)
               if (opt->lb[i] < opt->ub[i] && nlopt_istiny(opt->ub[i] - opt->lb[i]))
                   opt->ub[i] = opt->lb[i];


api/options.c
```diff
@@ -270,7 +270,7 @@ nlopt_result
 NLOPT_STDCALL nlopt_set_lower_bounds(nlopt_opt opt, const double *lb)
 {
      nlopt_unset_errmsg(opt);
-     if (opt && (opt->n == 0 || lb)) {
+     if (opt && (opt->n > 0) && lb) {
```
stevengj: This isn't right, because it will return `NLOPT_INVALID_ARGS` if `n == 0`, when actually that is not an error.

The code here is actually correct (it only allows a NULL `lb` in cases where `n == 0`, in which case nothing is written), so all we need to do is to silence the warning.   Couldn't that be accomplished by doing `lb && memcpy(...)` in place of `memcpy(...)`?

jschueller: ok, I did not read that correctly, maybe with two branches it will be more readable

jschueller: oh, then only the set methods are affecteor, or this can be done with just an "if"

stevengj: Either `opt->n > 0 && memcpy(...)` or `if (opt->n > 0) memcpy(...)` are fine (they are equivalent).

jschueller: it shoud be opt->lb as I guess that's what the static analysis is based on

stevengj: The point is that you can skip the `memcpy` whenever `opt->n == 0`, regardless of whether the pointers are non-NULL, so you might as well check `n > 0` before `memcpy` if you are going to insert a check here.

(Leave  the original`if (opt && (opt->n == 0 || lb)) {` as-is, since it was correct.)

jschueller: that's what I've done, except I have to specify that the condition depends on opt->lb as the static analysis is not smart enough to understand that opt-n>0 and opt->lb being valid are codependent events.

api/options.c
```diff
@@ -272,7 +272,8 @@ NLOPT_STDCALL nlopt_set_lower_bounds(nlopt_opt opt, const double *lb)
      nlopt_unset_errmsg(opt);
      if (opt && (opt->n == 0 || lb)) {
 	  unsigned int i;
-	  memcpy(opt->lb, lb, sizeof(double) * (opt->n));
+         if (opt->lb)
```
stevengj: It seems a lot clearer to me to check `opt->n > 0`, and more reliable too: it is possible in the future that we might have a non-NULL pointer here even when `n == 0`, and then you would get a warning if `lb == 0`.

jschueller: ok

PR #139, opened 2017-08-29
Update doc

jschueller:


jschueller:
I've got an error when generating the doc calling mkdocs; maybe I should figure it out first.

PR #140, opened 2017-09-09
Fix dllexport and dllimport for Windows

rickertm:
This patch fixes the DLL exports and imports for Windows. If a shared library is built, NLOPT_DLL must be defined for the library itself and everything linking against it (PUBLIC) whereas NLOPT_DLL_EXPORT must only be defined for the library itself (PRIVATE).


stevengj:
LGTM.  The travis failure is unrelated.

jschueller:
@rickertm did you encounter a failure with the previous code? it seems this was working as we got tests on windows. Or maybe your use-case is worth adding to the test-suite.

rickertm:
@jschueller Yes, due to the INTERFACE definition, no symbols are exported when using the default settings with BUILD_SHARED_LIBS=ON and nlopt.lib is not created. This leads to a fatal error when trying to link against it (tested on both Visual Studio 2015 and 2017).
The appveyor configuration uses BUILD_SHARED_LIBS=OFF which only builds a static library where exported symbols are not required.

PR #141, opened 2017-09-10
Add postfixes for debug and static builds on Windows

rickertm:
This patch adds a "d" postfix for debug builds (e.g., nloptd.lib and nloptd.dll) on Windows to enable the binaries for both debug and release builds to be installed at the same time as supported by the exported targets. It also adds a "s"/"sd" postfix for static builds (e.g., nlopts.lib and nloptsd.lib) to distinguish from shared builds.


jschueller:
-1, you can use CMAKE_<CONFIG>_POSTFIX to do exactly that: ie it should not be needed to do that inside the project

PR #146, opened 2017-09-11
Install PDB file with debug information when using Visual Studio

rickertm:
This patch installs the PDB file with debug symbols generated by MSVC into the installation's bin directory when building shared libraries and using the Debug or RelWithDebInfo configuration. This is only supported by CMake >= 3.1.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -259,6 +259,10 @@ install ( TARGETS ${nlopt_lib}
           ARCHIVE DESTINATION ${RELATIVE_INSTALL_LIB_DIR}
         )
 
+if (MSVC AND BUILD_SHARED_LIBS AND NOT CMAKE_VERSION VERSION_LESS 3.1)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -259,6 +259,10 @@ install ( TARGETS ${nlopt_lib}
           ARCHIVE DESTINATION ${RELATIVE_INSTALL_LIB_DIR}
         )
 
+if (MSVC AND BUILD_SHARED_LIBS AND ${CMAKE_MAJOR_VERSION} GREATER 2 AND ${CMAKE_MINOR_VERSION} GREATER 0)
```
jschueller: rather do ```CMAKE_VERSION VERSION_GREATER 3.1```

rickertm: The minimum version is 3.1, I think `NOT CMAKE_VERSION VERSION_LESS 3.1` would be better in this case?

jschueller: Indeed, VERSION_LESS is the strict comparison

PR #150, opened 2017-09-25
Change install destination for headers to relative path

rickertm:
Change to RELATIVE_INSTALL_INCLUDE_DIR similar to other install() variables as absolute paths are not accepted by CPack for an NSIS installer.


stevengj:
Is `RELATIVE_INSTALL_INCLUDE_DIR` a standard thing for cmake?  I can't find it documented anywhere.

rickertm:
With GNUInstallDirs the recommended relative path is in variable CMAKE_INSTALL_INCLUDEDIR. There are also CMAKE_INSTALL_FULL_<dir> variables provided by GNUInstallDirs in case this is required.
The RELATIVE_INSTALL_INCLUDE_DIR comes from the commands in lines 48-55 to convert the optionally specified variables into absolute ones.

jschueller:
This should totally be relative, thanks.

PR #152, opened 2017-10-13
Fixed Matlab find_package issue

hokiedsp:
To link Matlab's C/C++ Matrix Library (libmx.dll in Windows), find_package(Matlab) line in CMakeLists.txt needs to specify the MX_LIBRARY component as required.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -298,7 +298,7 @@ if (NLOPT_OCTAVE)
 endif ()
 
 if (NLOPT_MATLAB)
-  find_package (Matlab)
+  find_package (Matlab COMPONENTS MX_LIBRARY REQUIRED)
 endif ()
 
 if (OCTAVE_FOUND OR Matlab_FOUND)


stevengj:
Thanks — seems to be breaking Travis, though?

hokiedsp:
As far as I can tell (I'm still very much a cmake novice) MX_LIBRARY component must be present (i.e., REQUIRED) for the linking to succeed. If not, cmake does not present libmx.lib to the linker. 

To support what you'd like to do (automatically turning NLOPT_MATLAB option OFF), does CMAKE have a TRY-CATCH type of mechanism to trap the error?

stevengj:
@jschueller, do you know how to get cmake to do the right thing here?

jschueller:
you should do 
```find_package (Matlab COMPONENTS MX_LIBRARY)```

then
```if (OCTAVE_FOUND OR Matlab_MX_LIBRARY)```

instead of
```if (OCTAVE_FOUND OR Matlab_Found)```

jschueller:
@hokiedsp could you provide some more information on the error ? logs ? matlab version ? compiler ? os ?

hokiedsp:
@jschueller - Confirmed success on your branch. Please pull your branch to finalize the fix. Thanks!

PR #156, opened 2017-10-18
[pkgconfig] Mark libm as private

jschueller:


stevengj:
What does this do?

jschueller:
This means libm is internally linked to libnlopt but programs linked to libnlopt do not need to link to libm,
see https://gentoobrowse.randomdan.homeip.net/packages/sci-libs/nlopt

stevengj:
Does that only work on systems whose runtime linkers automatically pull in shared-library dependencies?  Or is pkgconfig smart enough to merge `Libs.private` with `Libs` if the system does not support "private" libs?

I guess that is most systems these days, but ancient Unices didn't support this feature.  Maybe we don't care about those anymore.

jschueller:
I may be too young to tell. But this seems pretty standard to wirte pkgconfig files this way.

PR #160, opened 2017-11-14
Add nlopt_get_numevals

jschueller:
We reuse stop.nevals which becomes stop.nevals_p, a pointer to opt.nevals, an int that we can return to the user.


stevengj:
LGTM, but maybe `get_numevals` would be a clearer name for the API (no need to change the internal name).

jschueller:
ok, I changed the name

mzaffalon:
The `set_numevals` method seems to be missing from the documentation, although it is implemented. Would you like a doc PR?

I came across it because I tried to reuse multiple times the same `nlopt_opt` object with `maxevals` and the optimization returned immediately because the `numevals` is not reset by `nlopt_optimize`, or so I understand. Out of curiosity: how would one have restarted the optimization before this PR?

What took me a while to figure it out was that from the Julia interface, the same opt object can be reused without additional action.

jschueller:
It should be there, see d64e335, maybe you're not reading the doc from master ?

jschueller:
oh, I dont think there should be a setter method, this is read-only.

jschueller:
Fixed

mzaffalon:
The set method is defined [here](https://github.com/stevengj/nlopt/pull/160/commits/12bb2a5ae111a5121f2ffc3af17c7699e6878c25#diff-79d6239b0fb190739e140c1fb31b75edR631).

My question is: how can one reuse the same nlopt_opt object if `maxevals` has been previously set but `numevals` cannot be reset?

jschueller:
Yes, I removed it, and reset numevals before running another optimization.

mzaffalon:
Is there a PR?

jschueller:
See 1226c1276dacf3687464c65eb165932281493a35, 10555232095b204267d01b88c7b47140687d82f8

PR #162, opened 2017-11-17
Simplify swig cmake logic

jschueller:


stevengj:
Will it be possible to ship the SWIG-generated files with the release tarball, so that users can install the Python code even if they don't have swig?

jschueller:
I think this will confuse users if we the source tarball is different from the release tarball.
Besides swig is easily installable these days.
From our side this makes the cmake logic quite complicated to support this, and we'd have to setup script to generate a specific release tarball.
So in regards to the drawbacks I'd say we rather not do it.

stevengj:
Yes, I guess most users of Python these days are using distros like Anaconda that include a swig package, or are on Linux where it is also easy.

(It would be nice to have a conda package for nlopt, too.)

PR #183, opened 2018-05-15
RFC: Per-variable relative tolerance criteria for x

aitap:
Dear Steven,

Thank you for your excellent library; my specialist's thesis would be much sadder without it.

I realized that I needed different relative tolerances for the variables I was optimizing (I wanted to be sure about plasma temperature and electron density, but had much less confidence in element concentrations), so I went on and implemented that.

Why you wouldn't want to merge this right away:

 - I had to make interface choices just to get the code to compile. I did my best to preserve existing interface, so the new functions are inconsistently named (`set_xtol_abs` and `set_xtol_abs1` vs `set_xtol_relv` and `set_xtol_rel`), some of them may return what can be considered lies (what should `double get_xtol_rel()` return when xtol_rel is a vector?), but the old behaviour is fully preserved (i.e. code using only `set_xtol_rel(double)` wouldn't notice any difference).
 - I had to change behaviour of some of the algorithms to make them compile again and to account for `stop->xtol_rel` being a vector. Some of the changes are nasty `FIXME`s which still work when all `xtol_rel` elements are same but may cause suboptimal behaviour otherwise. I would need some help to decide, for example, whether it's appropriate to allocate memory for temporary copy of `xtol_rel` to multiply it by `10`.
 - No attempt was made to make the new functions available from F77 or C++ before there's a decision on the interface for them.

If you are interested, can we try to work out the rest of the changes I need to implement for this to be mergeable?


stevengj:
I'm a bit skeptical about this.   My preferred change would be something like:

* Keep the relative tolerance `reltol` a dimensionless scalar with the same API, but make it apply to the whole vector: the stopping criterion should be ‖Δx‖ ≤ reltol × ‖x‖, where ‖⋯‖ is some norm (probably the L1 norm for computational simplicity).   This is better than the current component-wise relative tolerance, which is difficult to satisfy if some components of `x` happen to be nearly zero.

* Add a new API to specify optional weights `w` for the components of `x`.   ‖x‖ will then compute the weighted L1 norm ∑wᵢ|xᵢ|.   (Internally, we may just rescale the variables, since several of the algorithms implicitly assume that the components of `x` have similar scalings/units.)

aitap:
Thank you for your reply!

I like the fact that your approach manages to preserve existing API and express what I need. I'll try to implement it without disturbing existing `relstop` too much, then add the weighting.

src/util/stop.c
```diff
@@ -29,6 +29,15 @@
 
 /* utility routines to implement the various stopping criteria */
 
+static double vector_norm(unsigned n, double *vec, double *w)
+{
+    unsigned i;
+    double ret;
```
stevengj: initialize `ret = 0`

src/util/stop.c
```diff
@@ -29,6 +29,15 @@
...
+    for (i = 0; i < n; i++)
+        ret += (w ? w[i] : 1) * fabs(vec[i]);   /* this is L1 norm */
```
stevengj: rather than have this conditional inside the loop, might be better to just have two loops
```c
if (w)
    for (i = 0; i < n; i++) ret += fabs(vec[i]);
else
    for (i = 0; i < n; i++) ret += w[i] * fabs(vec[i]);
```

src/util/stop.c
```diff
@@ -49,6 +58,7 @@ int nlopt_stop_f(const nlopt_stopping * s, double f, double oldf)
 int nlopt_stop_x(const nlopt_stopping * s, const double *x, const double *oldx)
 {
     unsigned i;
+    /* TODO: iterative calculation of ‖Δx‖ */
```
stevengj: I guess you'll want something like a `norm_diff(n, x, oldx, w)` function that computes the weighted L1 norm of `x-oldx`.

src/util/stop.c
```diff
@@ -49,6 +75,8 @@ int nlopt_stop_f(const nlopt_stopping * s, double f, double oldf)
 int nlopt_stop_x(const nlopt_stopping * s, const double *x, const double *oldx)
 {
     unsigned i;
+    if (diff_norm(s->n, x, oldx, NULL) <= s->xtol_rel * diff_norm(s->n, x, oldx, NULL))
+        return 1;
     for (i = 0; i < s->n; ++i)
         if (!relstop(oldx[i], x[i], s->xtol_rel, s->xtol_abs[i]))
             return 0;
```
stevengj: I was thinking we would just pass `0` instead of `s->xtol_rel` here, i.e. only have the norm version of the `xtol_rel` test`.

src/util/stop.c
```diff
@@ -58,8 +86,10 @@ int nlopt_stop_x(const nlopt_stopping * s, const double *x, const double *oldx)
 int nlopt_stop_dx(const nlopt_stopping * s, const double *x, const double *dx)
 {
     unsigned i;
+    if (vector_norm(s->n, dx, NULL) <= s->xtol_rel * vector_norm(s->n, x, NULL))
+        return 1;
     for (i = 0; i < s->n; ++i)
-        if (!relstop(x[i] - dx[i], x[i], s->xtol_rel, s->xtol_abs[i]))
+        if (!relstop(x[i] - dx[i], x[i], 0 /* may be not a good idea */ , s->xtol_abs[i]))
```
stevengj: My thinking is that the xtol_rel test should only be the norm version, so passing 0 here is the right choice.

src/util/stop.c
stevengj: More simply, we could just test `if (fabs(x[i] - oldx[i]) > s->xtol_abs[i])`

src/util/stop.c
stevengj: More simply, we could just test `if (fabs(dx[i]) > s->xtol_abs[i])`

src/util/stop.c
```diff
@@ -49,17 +75,21 @@ int nlopt_stop_f(const nlopt_stopping * s, double f, double oldf)
 int nlopt_stop_x(const nlopt_stopping * s, const double *x, const double *oldx)
 {
     unsigned i;
+    if (diff_norm(s->n, x, oldx, s->x_weights) <= s->xtol_rel * diff_norm(s->n, x, oldx, s->x_weights))
```
stevengj: You are comparing `diff_norm < xtol_rel * diff_norm`, which seems wrong — it is equivalent to checking `1 <= xtol_rel`, no?

Should probably be:
```c
if (diff_norm(s->n, x, oldx, s->x_weights) <= s->xtol_rel * vector_norm(s->n, x, s->x_weights))
```
?

src/util/stop.c
```diff
@@ -74,6 +104,7 @@ static double sc(double x, double smin, double smax)
 int nlopt_stop_xs(const nlopt_stopping * s, const double *xs, const double *oldxs, const double *scale_min, const double *scale_max)
 {
     unsigned i;
+    /* TODO: scale ‖Δx‖... but not with w */
```
stevengj: Yeah, it seems like we'd need another variant of `vector_norm` and `diff_norm` that takes `scale_min` and `scale_max` arguments.  Or simply add these parameters and pass `NULL` in cases where they aren't used.

aitap:
Sorry to produce so many mistakes in the code I submit here.

I modified `vector_norm` and `diff_norm` to accept optional scaling arguments - is it still okay to handle 2*2 slightly different cases in 2 nested `if` clauses?

Couldn't help but notice that `nlopt_stop_x` and `nlopt_stop_dx` return 0 if a single `dx[i]` is `> xtol_abs[i]` while `nlopt_stop_xs` prefers to return 1 if any of the `xs[i]` fits the absolute stopping criterion. Is this intended?

stevengj:
> Couldn't help but notice that nlopt_stop_x and nlopt_stop_dx return 0 if a single dx[i] is > xtol_abs[i] while nlopt_stop_xs prefers to return 1 if any of the xs[i] fits the absolute stopping criterion. Is this intended?

That seems like a bug.

aitap:
Thank you for clarifying the intent of `nlopt_stop_xs`! I have made the appropriate changes, rebased the everything on top of current master and tested resulting `test/testopt` for undefined behaviour (using `-fsanitize=...`) with relative tolerance on x (`-x`) enabled.

What else needs to be done here?

src/api/options.c
```diff
@@ -157,10 +163,14 @@ nlopt_opt NLOPT_STDCALL nlopt_copy(const nlopt_opt opt)
             nopt->xtol_abs = (double *) malloc(sizeof(double) * (opt->n));
             if (!opt->xtol_abs)
                 goto oom;
+            nopt->x_weights = (double *) malloc(sizeof(double) * (opt->n));
```
stevengj: I'd rather have these initialized to NULL unless they are actually used — i.e. in the common case where the weights are equal they are not allocated.  But I guess we should do the same thing for `xtol_abs` as well, so that could be done in a separate PR.

src/api/options.c
```diff
@@ -628,6 +638,34 @@ nlopt_result NLOPT_STDCALL nlopt_get_xtol_abs(const nlopt_opt opt, double *xtol_
...
+            opt->x_weights[i] = x_weights;
+        return NLOPT_SUCCESS;
+    }
+    return NLOPT_INVALID_ARGS;
+}
+
+nlopt_result NLOPT_STDCALL nlopt_get_x_weights(const nlopt_opt opt, double *x_weights)
+{
+    memcpy(x_weights, opt->x_weights, opt->n * sizeof(double));
+    return NLOPT_SUCCESS;
```
stevengj: This should have a couple of argument checks, similar to `nlopt_get_lower_bounds`

stevengj:
It will also need an update to the documentation.

src/api/options.c
```diff
@@ -628,6 +636,56 @@ nlopt_result NLOPT_STDCALL nlopt_get_xtol_abs(const nlopt_opt opt, double *xtol_
...
+{
+    if (opt) {
+        nlopt_unset_errmsg(opt);
+        if (!opt->x_weights) {
+            opt->x_weights = (double *) calloc(opt->n, sizeof(double));
+            if (!opt->x_weights) {
+                return NLOPT_OUT_OF_MEMORY;
+            }
+        }
+        memcpy(opt->x_weights, x_weights, opt->n * sizeof(double));
```
stevengj: It should probably check that all of the weights are > 0 or return an INVALID_ARGS

src/api/options.c
```diff
@@ -640,6 +640,11 @@ nlopt_result NLOPT_STDCALL nlopt_set_x_weights(nlopt_opt opt, const double *x_we
 {
     if (opt) {
         nlopt_unset_errmsg(opt);
+        unsigned i;
+        for (i = 0; i < opt->n; i++) {
+            if (x_weights[i] <= 0)
+                return NLOPT_INVALID_ARGS;
```
stevengj: `return ERR(NLOPT_INVALID_ARGS, opt, "invalid negative weight");`

src/api/options.c
```diff
@@ -654,7 +659,7 @@ nlopt_result NLOPT_STDCALL nlopt_set_x_weights(nlopt_opt opt, const double *x_we
 
 nlopt_result NLOPT_STDCALL nlopt_set_x_weights1(nlopt_opt opt, double x_weights)
 {
-    if (opt) {
+    if (opt && x_weights > 0) {
```
stevengj: Better to have
```c
if (x_weights < 0) return ERR(NLOPT_INVALID_ARGS, opt, "invalid negative weight");
```
so that there is an informative error message.

aitap: Should we make `x_weights == 0` an error, too? I am assuming we should, since there already is `xtol_rel = 0` to disable the stopping criterion.

stevengj:
Looks good, I'll merge once tests are green.  Thanks for sticking with this!

PR #189, opened 2018-06-02
Octave 4.4 compatibilty

badshah400:
nlopt fails to build when compiled against octave 4.4. This is because of the following two reasons:
* `octave-config` no longer accepts the `PREFIX` keyword, the `OCTAVE_HOME` keyword is supposed to be used instead. See [this changeset](https://hg.savannah.gnu.org/hgweb/octave/rev/b95c430c0649).
* `do_multi_index_op()` has been dropped with `octave::feval()` being the recommended way to achieve the same output. See [here](https://hg.savannah.gnu.org/hgweb/octave/rev/c6714ae1c06c).

This PR fixes these issues and makes nlopt compatible with Octave 4.4.


badshah400:
Thanks for the quick review and merge.

PR #197, opened 2018-07-26
Compile tests only in top-level projects.

jdumas:
This is a small change, which basically means that tests won't be compiled when `nlopt` is used in another project via `add_subdirectory()`. Please let me know if you have any question.


jschueller:
an option to disable building of tests would be more generic

stevengj:
Are the tests causing a problem?   I agree in principle that there should be a general mechanism (e.g. `-DNLopt_tests=Off`) to disable building the tests, but they are so small it's not clear it's worth the trouble.

(I just encountered a problem building the tests with a cross compiler because they weren't being linked by the C++ linker, but that is hopefully fixed by 6e7f31fefe4398fafb04a006db4bbf834516d3d1.)

jdumas:
I'm building my own code against nlopt, and I don't have to have both set of tests (the ones for nlopt and mine) in the same project, since I'm only interested in my own tests. And when a test in nlopt fails it shouldn't impact my own tests (e.g. right now I have an issue with the guile test that is not compiling).

I can add an option if you prefer, and set its default value based on whether nlopt is a top-level project or not.

jdumas:
Ok done!

CMakeLists.txt
```diff
@@ -31,6 +31,12 @@ option (NLOPT_GUILE "build guile bindings" ON)
 option (NLOPT_SWIG "use SWIG to build bindings" ON)
 option (NLOPT_LINK_PYTHON "link Python libs" ON)
 
+if (CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)
+  option (NLOPT_BUILD_TESTS "build unit tests" ON)
```
stevengj: Probably just `NLOPT_TESTS` to be shorter

stevengj:
LGTM, thanks!

PR #198, opened 2018-07-26
Enable CXX by default

jschueller:
It's silly not to provide all algorithms by default,
c++ compiler are easy to find these days, worst case it can still be disabled

Also remove lib suffix

### DIFF
--- a/doc/docs/NLopt_Installation.md
+++ b/doc/docs/NLopt_Installation.md
@@ -142,14 +142,13 @@ Note, however, that if you do this then Guile may not know where to load the `nl
 NLopt with C++ algorithms
 -------------------------
 
-NLopt, as-is, is callable from C, C++, and Fortran, with optional Matlab and GNU Octave plugins (and even installs an `nlopt.hpp` C++ header file to allow you to call it in a more C++ style). By default, it includes only subroutines written in C (or written in Fortran and converted to C), to simplify linking. If you configure with:
+NLopt, as-is, is callable from C, C++, and Fortran, with optional Matlab and GNU Octave plugins (and even installs an `nlopt.hpp` C++ header file to allow you to call it in a more C++ style). By default, it includes subroutines written in C (or written in Fortran and converted to C) and C++. If you configure with:
 
 ```sh
-cmake -DNLOPT_CXX=ON ..
+cmake -DNLOPT_CXX=OFF ..
 ```
 
-however, it will also include algorithms implemented in C++ (currently, just the StoGO algorithm), and the resulting library will be called `libnlopt_cxx` and is linked with `-lnlopt_cxx`.
+however, it will disable algorithms implemented in C++ (StoGO and AGS algorithms).
 
-The `libnlopt_cxx` has the *same* interface as the ordinary NLopt library, and can *still* be called from ordinary C and Fortran programs. However, to use it you must also *link* with the C++ standard libraries. The easiest way to do this is to link with the C++ linker: compile your source files into `.o` object files, and then call the C++ compiler to link these `.o` files with `-lnlopt_cxx` into your executable program.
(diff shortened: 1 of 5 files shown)


stevengj:
https://github.com/stevengj/nlopt/blob/master/doc/docs/NLopt_Installation.md also needs to be updated.

Does the Matlab plugin work with the C++ version?  The mex file may need to end in `.cpp` for Matlab to use the C++ linker.

jschueller:
I think cmake has what it takes to link to c++ when nlopt_lib is c++ enabled, but I did not test yet.

doc/docs/NLopt_Installation.md
```diff
@@ -142,14 +142,13 @@ Note, however, that if you do this then Guile may not know where to load the `nl
 NLopt with C++ algorithms
 -------------------------
 
-NLopt, as-is, is callable from C, C++, and Fortran, with optional Matlab and GNU Octave plugins (and even installs an `nlopt.hpp` C++ header file to allow you to call it in a more C++ style). By default, it includes only subroutines written in C (or written in Fortran and converted to C), to simplify linking. If you configure with:
+NLopt, as-is, is callable from C, C++, and Fortran, with optional Matlab and GNU Octave plugins (and even installs an `nlopt.hpp` C++ header file to allow you to call it in a more C++ style). By default, it includes subroutines written in C (or written in Fortran and converted to C) and as well in C++. If you configure with:
```
stevengj: "and as well" is redundant, and "in" is misplace.   Just say "and C++"

doc/docs/NLopt_Installation.md
```diff
@@ -142,14 +142,13 @@ Note, however, that if you do this then Guile may not know where to load the `nl
...
 ```sh
-cmake -DNLOPT_CXX=ON ..
+cmake -DNLOPT_CXX=OFF ..
 ```
 
-however, it will also include algorithms implemented in C++ (currently, just the StoGO algorithm), and the resulting library will be called `libnlopt_cxx` and is linked with `-lnlopt_cxx`.
+however, it will disable algorithms implemented in C++ (StoGO and AGS algorithms).
 
-The `libnlopt_cxx` has the *same* interface as the ordinary NLopt library, and can *still* be called from ordinary C and Fortran programs. However, to use it you must also *link* with the C++ standard libraries. The easiest way to do this is to link with the C++ linker: compile your source files into `.o` object files, and then call the C++ compiler to link these `.o` files with `-lnlopt_cxx` into your executable program.
+The resulting library has the *same* interface as the ordinary NLopt library, and can *still* be called from ordinary C, C++, and Fortran programs. However, to use it doesnt have to *link* with the C++ standard libraries.
```
stevengj: "However, one no longer has to link with..."

stevengj:
Not having this on by default was never about C++ compilers being hard to find.  The problems were:

* There are too many C++ compilers, and they are often incompatible with one another — the pure C NLopt library works with any C++ compiler, but the C++ NLopt may only work with the C++ compiler it was built with.    But these days there is a lot less diversity in standard libraries.   

* Callers from Fortran and C may have to know to link with `-lstdc++`, but that situation is somewhat improved these days because many systems automatically recognize shared-library dependencies.

stevengj:
(It used to be that Matlab only shipped with `lcc` on Windows, so out-of-the-box it only worked with pure-C plugins.  I'm not sure what the situation is these days.)

jschueller:
I just tested with matlab; it builds and loads fine, I guess it's fine on windows too at least when using cmake.

jschueller:
I also updated the doc.

stevengj:
Why was this closed?

jschueller:
I thought you were not too found of this change.

stevengj:
I have mixed feelings, but I’m inclined to think that the pros outweigh the cons nowadays thanks to .so dependencies.

PR #201, opened 2018-07-27
Set PATH on windows hosts only

jschueller:
With nmake there is no subdir, so find the nlopt dll here too.
Set it only for windows hosts, not accurate when cross-compiling.


stevengj:
Failing AppVeyor...

PR #228, opened 2018-10-24
Add -Wall and -Wextra, fix lots of warnings

### LINKED ISSUE #226: Lots of C/C++ warnings ...
I'm compiling nlopt within an embedded project which has -Wall -Wextra turned on by default and get lots of warnings:

- warning: variable 'xxxx' set but not used [-Wunused-but-set-variable]
- deprecated declaration in f77api.c

I made myself ready to fix them, but now I'm unsure if there is some intention behind these unused variables? If they need to stay in I'd suggest to use `(void) XXX;` to silence the warnings, otherwise I'd remove them, please advise.
For the deprecated functions in f77api.c and f77funcs.h I'd add a GCC pragma (with correct ifdefs) to turn the warning off for these files.

taliesin:
- remove set but not used variables
 - (void) cast unused function arguments
 - create internal function duplicates for some actually
   deprecated functions, for the f77 files disable
   the deprecation warning completely.

I'm not sure if the changes in CMakeLists.txt are appropriate as I'm not much of a cmaker, please advise.

Closes #226.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,9 +78,6 @@ if(POLICY CMP0042)
   cmake_policy(SET CMP0042 NEW)
 endif()
 
-set (CMAKE_C_FLAGS "-Wall -Wextra ${CMAKE_C_FLAGS}")
-set (CMAKE_CXX_FLAGS "-Wall -Wextra ${CMAKE_CXX_FLAGS}")
-
 include (CheckIncludeFiles)
 include (CheckFunctionExists)
 include (CheckTypeSize)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,6 +78,9 @@ if(POLICY CMP0042)
   cmake_policy(SET CMP0042 NEW)
 endif()
 
+set (CMAKE_C_FLAGS "-Wall -Wextra ${CMAKE_C_FLAGS}")
+set (CMAKE_CXX_FLAGS "-Wall -Wextra ${CMAKE_CXX_FLAGS}")
+
 include (CheckIncludeFiles)
 include (CheckFunctionExists)
 include (CheckTypeSize)
--- a/src/algs/ags/solver.cc
+++ b/src/algs/ags/solver.cc
@@ -58,7 +58,7 @@ namespace
           right[i] = mRightBound[i];
         }
       }
-      int GetOptimumPoint(double* y) const {return 0;}
+      int GetOptimumPoint(double* y) const {(void) y; return 0;}
       double GetOptimumValue() const {return 0;}
     };
 }
--- a/src/algs/ags/solver.cc
+++ b/src/algs/ags/solver.cc
@@ -58,7 +58,7 @@ namespace
           right[i] = mRightBound[i];
         }
       }
-      int GetOptimumPoint(double* y) const {(void) y; return 0;}
+      int GetOptimumPoint(double*) const {return 0;}
       double GetOptimumValue() const {return 0;}
     };
 }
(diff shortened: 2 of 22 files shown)


CMakeLists.txt
```diff
@@ -78,6 +78,9 @@ if(POLICY CMP0042)
   cmake_policy(SET CMP0042 NEW)
 endif()
 
+set (CMAKE_C_FLAGS "-Wall -Wextra ${CMAKE_C_FLAGS}")
+set (CMAKE_CXX_FLAGS "-Wall -Wextra ${CMAKE_CXX_FLAGS}")
```
jschueller: -1
These should be set outside of cmake, you can set these in the travis script for example, even with Werror if you want.
It's sensitive to the plaftorm/compiler. Even gcc won't yield the same warnings dependings on the version.

src/algs/ags/solver.cc
```diff
@@ -58,7 +58,7 @@ namespace
           right[i] = mRightBound[i];
         }
       }
-      int GetOptimumPoint(double* y) const {return 0;}
+      int GetOptimumPoint(double* y) const {(void) y; return 0;}
```
jschueller: I think you can simply use `GetOptimumPoint(double* ) const {return 0;}`

taliesin: for C++ this is correct, but in C that won't work (for each compiler), I'll change it here.

CMakeLists.txt
taliesin: If I change it only in travis, this will not build with these flags on the local machine using the suggested build method:
* mkdir build; cd build
* cmake ..
* make
Is there a better way?

taliesin:
Could you please also comment on the AppVeyor build fail, I don't comprehend what's wrong?
__cplusplus not found?

jschueller:
Yes, that's because msvc doesnt support these cxx flags.

CMakeLists.txt
jschueller: These should not be fixed into the CMakeLists, I think it's better to let each one choose his compiler flags.

src/api/nlopt.h
```diff
@@ -288,6 +288,7 @@ NLOPT_EXTERN(void) nlopt_munge_data(nlopt_opt opt, nlopt_munge2 munge, void *dat
    for code that uses a deprecated function */
 #if defined(__GNUC__) && (__GNUC__ > 3 || (__GNUC__==3 && __GNUC_MINOR__ > 0))
 #  define NLOPT_DEPRECATED __attribute__((deprecated))
+#  define COMPILER_PROVIDES_DEPRECATED_ATTRIBUTE
```
stevengj: We shouldn't define this in `nlopt.h` because that exports the symbol to callers.  Probably simplest to just check `defined(__GNUC__) && (__GNUC__ > 3 || (__GNUC__==3 && __GNUC_MINOR__ > 0))` since you only use this in two places.

CMakeLists.txt
taliesin: Ok, I'll put it in the travis file:

`cmake -DCMAKE_INSTALL_PREFIX=~/.local -DNLOPT_MATLAB=OFF -DCMAKE_C_FLAGS='-Wall -Wextra -Werror' -DCMAKE_CXX_FLAGS='-Wall -Wextra -Werror' ..`

.. but only in the first line (not cluttering the file too much) as the warning should fail there already and the mingw builds should not do anything differently, I suppose.

I'll add a commit to the current branch with the fixes. If you like to have a cleaner history I'll squash the commits in a new branch and you may merge from there?

taliesin:
Ok, now the swig stuff produces lots of warnings, which I can't fix, because it's in swig code.
I'll squash the fixes into one commit for version 2 and leave it at that for now.

PR #230, opened 2018-10-25
Add -Wall and -Wextra, fix lots of warnings

### LINKED ISSUE #226: Lots of C/C++ warnings ...

taliesin:
- remove set but not used variables
 - (void) cast unused function arguments
 - create internal function duplicates for some actually
   deprecated functions, for the f77 files disable
   the deprecation warning completely.
 - add warning flags to travis

Note: some of the maybe-unitialized warnings do not
  show in gcc 7.3.0, but travis seems to use a different one.

This closes #226.

### DIFF
--- a/src/api/nlopt.h
+++ b/src/api/nlopt.h
@@ -301,25 +301,11 @@ NLOPT_EXTERN(nlopt_result) nlopt_minimize(nlopt_algorithm algorithm, int n, nlop
                                           double *minf, /* out: minimum */
                                           double minf_max, double ftol_rel, double ftol_abs, double xtol_rel, const double *xtol_abs, int maxeval, double maxtime) NLOPT_DEPRECATED;
 
-/* this is still used internally by a few algorithms, it's the same as nlopt_minimize, but without the deprecation */
-NLOPT_EXTERN(nlopt_result) internal_nlopt_minimize(nlopt_algorithm algorithm, int n, nlopt_func_old f, void *f_data,
-                                          const double *lb, const double *ub, /* bounds */
-                                          double *x,    /* in: initial guess, out: minimizer */
-                                          double *minf, /* out: minimum */
-                                          double minf_max, double ftol_rel, double ftol_abs, double xtol_rel, const double *xtol_abs, int maxeval, double maxtime);
-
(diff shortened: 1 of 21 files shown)


taliesin:
Anything else to be done to have that merged?

src/api/nlopt.h
```diff
@@ -301,12 +301,26 @@ NLOPT_EXTERN(nlopt_result) nlopt_minimize(nlopt_algorithm algorithm, int n, nlop
                                           double *minf, /* out: minimum */
                                           double minf_max, double ftol_rel, double ftol_abs, double xtol_rel, const double *xtol_abs, int maxeval, double maxtime) NLOPT_DEPRECATED;
 
+/* this is still used internally by a few algorithms, it's the same as nlopt_minimize, but without the deprecation */
+NLOPT_EXTERN(nlopt_result) internal_nlopt_minimize(nlopt_algorithm algorithm, int n, nlopt_func_old f, void *f_data,
+                                          const double *lb, const double *ub, /* bounds */
+                                          double *x,    /* in: initial guess, out: minimizer */
+                                          double *minf, /* out: minimum */
+                                          double minf_max, double ftol_rel, double ftol_abs, double xtol_rel, const double *xtol_abs, int maxeval, double maxtime);
+
```
stevengj: I think it would be better to just update the algorithms to use the new API.

(We shouldn't export an `internal_*` function in `nlopt.h` in any case.)

stevengj:
rebased

src/api/nlopt.h
stevengj: Updated to eliminate these calls.

PR #242, opened 2018-12-10
AUGLAG: Fix arithmetic exception

### LINKED ISSUE #235: auglag divide by zero
On line 177 of auglag.c there is:
```
d.rho = MAX(1e-6, MIN(10, 2 * fabs(*minf) / con2));
```
However, if there are only inequality constraints and they are all satisfied then con2 == 0. Something like 
```
if (con2 == 0) d.rho = 10;
else d.rho = MAX(1e-6, MIN(10, 2 * fabs(*minf) / con2));
```
would be a simple fix, but I'm just guessing.

jschueller:
Fixes #235


stevengj:
Thanks!

PR #250, opened 2019-03-15
Export Fortran symbols

### LINKED ISSUE #249: Fortran API is not exported on Windows
I compile the lib with cmake and msvc, the resulting lib works well when call from a C program, but can not call from a fortran program, seems that it does not export fortran api `nlo_*`.

jschueller:
Fixes #249

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,13 +13,10 @@ matrix:
           - guile-2.0-dev
           - liboctave-dev
           - cmake
-          - g++-mingw-w64-i686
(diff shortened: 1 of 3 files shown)


jschueller:
maybe a bugfix version (2.5.1 ?) should be released sometime before the summer ?

stevengj:
Thanks!

stevengj:
Probably needs to be 2.6.0 because of #198.

PR #260, opened 2019-04-12
Fix c89 build

jschueller:


stevengj:
Closing in favor of #261; sorry, we both worked on this at the same time.

PR #261, opened 2019-04-12
c89 compatibility

### LINKED ISSUE #259: DIRserial.c(29): error C2275: 'integer' : illegal use of this type as an expression
it seems nlopt master does not build anymore with vs2010 since #230:
```src\algs\direct\DIRserial.c(29): error C2275: 'integer' : illegal use of this type as an expression```
maybe it's the void cast syntax ?

stevengj:
Fixes #259, adds `-std=c89 -pedantic -Werror` to compilation flags on Travis to catch future C89 errors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ matrix:
         - pip install mkdocs python-markdown-math --user
         - PATH=$PATH:~/.local/bin mkdocs build
         - mkdir build && pushd build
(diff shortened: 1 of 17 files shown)


jschueller:
can you include my appveyor changes ?

jschueller:
maybe it's not necessary to build vs14 twice for 32 and 64 bits

stevengj:
I would prefer to test both win32 and win64 with the same VS version.

stevengj:
Changed the test to be a little less strict and allow POSIX 2001 functions, so that it allows functions like `vsnprintf` and `isinf` that are widely available even on C89-ish compilers.

vhaguiar:
bobyqa seems unstable, given the same initial value, algorithm is deterministic, so it should be stable.

PR #264, opened 2019-04-18
Add algorithm enum string conversion

jschueller:
I'm not sure about the methods names, nor if it should be methods attached to nlopt_opt.


stevengj:
What's the motivation for this feature?

jschueller:
I want to be able to identify an algorithm by it's string, and to retrieve the available names in nlopt.
Currently I'm doing this in the c++ client side, but I found out Ceres Solver also has these string conversion utils built-in, and I found it was neat to have in nlopt as well.

stevengj:
Why do you want to identify an algorithm from its string? Some kind of UI thing, I guess, e.g. the user specifies the algorithm in an input file that you parse?

(I’m not opposed to the feature, I just want to be clear on the application.)

src/api/nlopt.h
```diff
@@ -207,6 +207,9 @@ NLOPT_EXTERN(unsigned) nlopt_get_dimension(const nlopt_opt opt);
 
 NLOPT_EXTERN(const char *) nlopt_get_errmsg(nlopt_opt opt);
 
+NLOPT_EXTERN(const char *) nlopt_algo_to_string(nlopt_algorithm algorithm);
+NLOPT_EXTERN(nlopt_algorithm) nlopt_algo_from_string(const char *name);
```
stevengj: Let’s write out “algorithm” here.

jschueller: ok

jschueller:
@stevengj I also added string converter for the result enum

src/api/options.c
```diff
@@ -839,4 +839,106 @@ const char *nlopt_get_errmsg(nlopt_opt opt)
...
+    if (strcmp (name, nlopt_algorithm_to_string(i)) == 0)
+      return i;
+  }
+  return -1;
+}
+
 /*************************************************************************/
+
+
+const char *nlopt_result_to_string(nlopt_result result)
```
stevengj: This function seems a little inconsistent with `nlopt_algorithm_to_string`, since `nlopt_result_to_string` gives a "human-readable" description and `nlopt_algorithm_to_string` gives the C enum.

This is more like `nlopt_algorithm_name` …

stevengj: Maybe it should be called `nlopt_result_name` (and delete `nlopt_result_from_string`)?

stevengj:
(The docs also need an update.)

jschueller:
ok, I stuck with a raw enum conversion like the algorithm, updated doc.

PR #265, opened 2019-04-23
add argument checks to nlopt_get_xtol_abs

aitap:
Similar `nlopt_get_*` functions in `options.c` always check for
`(opt && (opt->n == 0 || argument))`, requiring a valid `nlopt * opt`
and only allowing a null argument when the optimizer is "empty".

`nlopt_get_xtol_abs` seems to have been left out (13e0f277).


stevengj:
Looks good, thanks!

PR #268, opened 2019-04-26
do not allocate xtol_abs unless needed

aitap:
As discussed in #183, it is beneficial to avoid allocating potentially
huge buffers of size `n` unless `x`-tolerance criteria are used.

This patch adds checks for `!opt->xtol_abs` and behaves as if
`xtol_abs[i] == 0 ∀i` if true.

### DIFF
--- a/src/algs/bobyqa/bobyqa.c
+++ b/src/algs/bobyqa/bobyqa.c
@@ -3123,9 +3123,11 @@ nlopt_result bobyqa(int n, int npt, double *x,
 
     /* SGJ, 2009: compute rhoend from NLopt stop info */
     rhoend = stop->xtol_rel * (rhobeg);
-    for (j = 0; j < n; ++j)
-	 if (rhoend < stop->xtol_abs[j] / fabs(s[j]))
-	      rhoend = stop->xtol_abs[j] / fabs(s[j]);
+    if (stop->xtol_abs) {
+        for (j = 0; j < n; ++j)
+         if (rhoend < stop->xtol_abs[j] / fabs(s[j]))
+              rhoend = stop->xtol_abs[j] / fabs(s[j]);
+    }
 
 
 /*     This subroutine seeks the least value of a function of many variables, */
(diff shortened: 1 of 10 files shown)


stevengj:
There are also a few other explicit references to `xtol_abs` that need to be fixed: in `bobyqa.c`, in `cdirect.c`, in `hybrid.c`, in `cobyla.c`, in `sbplx.c`, in `newuoa.c`, in `praxis.c`, in `stogo/local.cc`, in `api/optimize.c`

aitap:
Of course, I should have read the code instead of relying on `make test` to produce memory errors.

There are parts in [cdirect.c](https://github.com/stevengj/nlopt/blob/master/src/algs/cdirect/cdirect.c#L382) and [hybrid.c](https://github.com/stevengj/nlopt/blob/master/src/algs/cdirect/hybrid.c#L152) where both stopping criteria on `x` seem to be checked manually, except the relative stopping criterion is
```c
dx[i] <= (ub[i] - lb[i]) * xtol_rel // ∀ i
// where lb and ub are 0 and 1 unless NOSCAL is used
```
instead of `dx[i] < xtol_rel * x[i] // ∀ i` or `||∆x|| < xtol_rel * ||wx||`.

If this is integral to the algorithm, I can preserve the current behaviour by adding checks:
```diff
-	  if (w[i] > (p->stop->xtol_abs[i] ) &&
+	  if (w[i] > (p->stop->xtol_abs ? p->stop->xtol_abs[i] : 0) &&
```

```diff
- && w[i] > p->stop->xtol_abs[i])
+ && w[i] > p->stop->xtol_abs ? p->stop->xtol_abs[i] : 0)
```
Is this the right thing to do, though?

stevengj:
@aitap, I think that code predates the code in `stop.c` — it can be replaced with calls to `stop.c` routines

aitap:
Sorry, my analysis was wrong. Those are not `dx`, but widths of hyper-rectangle sides. The `(ub[i] - lb[i]) * xtol_rel` criterion makes much more sense now.

I tried to avoid branching in for-loops, but in some cases letting a ternary operator in led to more comprehensible code.

stevengj:
LGTM, thanks!

Issue #270, opened 2019-05-03
Matlab 2017b - Ubuntu 18.04 - Tutorial fails


NicolasRiel:
Hi everyone,

I was able to install NLopt with matlab 2017b using this command:

`sudo cmake .. -DMatlab_ROOT_DIR=/home/seph/MATLAB/R2017b/`

However, when I try to run the example from the tutorial I get the following error message.  

I was wondering if anyone had an idea on how to fix it? 

Cheers!

_Invalid MEX-file '/home/seph/MATLAB/R2017b/bin/nlopt_optimize.mexa64':
Missing symbol '_ZNSt13runtime_errorC1EPKc' required by '/usr/local/lib/libnlopt.so.0->/home/seph/MATLAB/R2017b/bin/nlopt_optimize.mexa64'
Missing symbol '_ZNSt13runtime_errorC1ERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE' required by
'/usr/local/lib/libnlopt.so.0->/home/seph/MATLAB/R2017b/bin/nlopt_optimize.mexa64'
Missing symbol '_ZNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEE10_M_replaceEmmPKcm' required by
'/usr/local/lib/libnlopt.so.0->/home/seph/MATLAB/R2017b/bin/nlopt_optimize.mexa64'
Missing symbol '_ZNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEE12_M_constructEmc' required by
'/usr/local/lib/libnlopt.so.0->/home/seph/MATLAB/R2017b/bin/nlopt_optimize.mexa64'
Missing symbol '_ZNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEE9_M_appendEPKcm' required by
'/usr/local/lib/libnlopt.so.0->/home/seph/MATLAB/R2017b/bin/nlopt_optimize.mexa64'
Missing symbol '_ZNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEE9_M_createERmm' required by
'/usr/local/lib/libnlopt.so.0->/home/seph/MATLAB/R2017b/bin/nlopt_optimize.mexa64'._

Here is the config log:

-- The C compiler identification is GNU 7.4.0
-- The CXX compiler identification is GNU 7.4.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
[...]

PR #272, opened 2019-05-08
Set matlab extension linker language

jschueller:
Also matlab's c++ library might be older and incompatible with the compiler's one,
so mention in the doc the possible workaround.

Added a test.


stevengj:
Closes #270?

jschueller:
yes

PR #294, opened 2019-09-18
Correct compilation error with C++17 (unary functions removed)

ocommowi:
When compiling the current master on Windows with Visual studio 2017 or higher with the C++17 standard, compilation errors show up due to unary_function being removed from the C++ standard from that point on.

The correction being very simple (just remove the inheritance from unary_function from C++ standard C++11), I propose here a little patch to handle this.

Tell me what you think

### DIFF
--- a/src/algs/stogo/tools.h
+++ b/src/algs/stogo/tools.h
@@ -36,7 +36,11 @@ public:
   friend ostream & operator << (ostream &, RCTrial) ;
 };
 
+#if __cplusplus < 201103L
 class TrialGT : public unary_function<Trial, bool>
+#else
+class TrialGT
+#endif
 // Predicate for Trial (needed for remove_if)
 {
 public:


ocommowi:
Just putting an additional comment to this. I realized just now that for the preprocess variable __cplusplus to have the right value, you need on windows to add /Zc:__cplusplus to your CMAKE_CXX_FLAGS . I had this already included in my cmake and forgot about it. May be adding something like below in some well chose CMakeLists.txt would be good to have this PR fully work

if (WIN32)
set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /Zc:__cplusplus"
endif()

jschueller:
shouldnt __cplusplus be already defined by the compiler ?

ocommowi:
On every compiler, yes, except visual studio which defines by default a fake __cplusplus set to 199711L even if you set the C++ standard to anything else. But the flag I mention enables __cplusplus to have the right value like in every other compiler. They describe it there: https://docs.microsoft.com/en-us/cpp/build/reference/zc-cplusplus

PR #295, opened 2019-09-18
Add usage of _MSVC_LANG for windows C++ standard check

ocommowi:
Following discussion on PR #294 , here is an improvement that should work for all compilers (windows or not). Sorry about the error in the first place


ocommowi:
Yes indeed, corrected

ocommowi:
Tell me if there is anything else I need to do to have this PR merged. That would help in our dev process. Thanks !

PR #312, opened 2020-01-16
MEX library symbols visibility set to default

nicolapiccinelli:
By default, the mexFunctions seems to be hidden in the newer Matlab version, this fix force the compiler to keep the symbols visible.

### DIFF
--- a/src/octave/CMakeLists.txt
+++ b/src/octave/CMakeLists.txt
@@ -6,7 +6,12 @@ if (Matlab_FOUND AND Matlab_MX_LIBRARY)
   cmake_minimum_required (VERSION 3.3) # for the matlab_add_mex macro
 
   matlab_add_mex (NAME nlopt_optimize-mex SRC nlopt_optimize-mex.c OUTPUT_NAME nlopt_optimize LINK_TO ${nlopt_lib})
-  set_target_properties(nlopt_optimize-mex PROPERTIES COMPILE_FLAGS "-fvisibility=default")
+  if (CMAKE_VERSION VERSION_LESS 3.14.0)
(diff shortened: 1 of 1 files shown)


jschueller:
it's not clear whether the bug is fixed in cmake or not for Linux
https://gitlab.kitware.com/cmake/cmake/issues/18391 seems to only fix it for Windows

nicolapiccinelli:
I will add the condition and update the pull request

stevengj:
We could also just add
```c
#ifdef __GNUC__
__attribute__ ((dllexport))
#endif
```
to the `mexFunction` declaration in https://github.com/stevengj/nlopt/blob/master/src/octave/nlopt_optimize-mex.c, no?

nicolapiccinelli:
I've tried and it works too, so It depends if you want to fix directly the code or only the configuration. But probably it's better to add both, for an extra security?

jschueller:
could you try if it it fixed in latest cmake version ?

nicolapiccinelli:
> could you try if it it fixed in latest cmake version ?

without any fixes?

jschueller:
yes

nicolapiccinelli:
With CMake 3.16.2 it works out of the box

nicolapiccinelli:
At this point, I think we can keep the fix only in the CMakeLists.txt and add a guard to guarantee only the backward compatibility. In Ubuntu 18.04 (the last LTS version) the CMake version shipped with the OS is the 3.10.2

nicolapiccinelli:
I've investigated and from the 3.14.0 it works out of the box. At this point, I think we can keep the fix only in the CMakeLists.txt and add a guard to guarantee only the backward compatibility. In Ubuntu 18.04 (the last LTS version) the CMake version shipped with the OS is the 3.10.2

Issue #118, opened 2017-04-08
COBYLA hangs


bluescarni:
COBYLA seems to hang on the following code:

https://gist.github.com/bluescarni/bdb1981c80966ebae5d92551b605347d

Compilation command line, on 64bit linux GCC 6.3 and NLopt 2.4.2:

```console
g++ -std=c++11 -O2 -DNDEBUG file.cpp -lnlopt -march=native
```

If the dimension of the problem is changed, from 150, to 149 or 151, the optimisation does not hang. If the dimension is 150, the optimisation seems to hang after ~170 objfun/fitness evaluations (strangely enough, the exact number is 171 on my Gentoo linux installation, and 175 on my Arch linux installation).

The optimisation problem is No. 5.1. from Luksan, L., and Jan Vlcek. “Sparse and partially separable test problems for unconstrained and equality constrained optimization”. It has only equality constraints.

ericjster:
For what it is worth, using NLOPT_AUGLAG with the local solver NLOPT_LN_NEWUOA or NLOPT_LN_BOBYQA, can solve this problem in about 28000 evaluations, in about 90 seconds on my machine, with ftol_rel = ftol_abs = 1e-6. I only mention this as a fellow user, if it might help to work around the infinite loop in COBYLA.

bluescarni:
@ericjster Thanks for the pointer! I thought it might be somehow related to the equality constraints as, as far as I understand, those are not handled by the original COBYLA algorithm...

bluescarni:
@ericjster btw, shameless plug, but if you might find our project pagmo interesting:

https://github.com/esa/pagmo2

I was writing the NLopt wrappers for pagmo when I ran into this issue.

ericjster:
Thanks Francesco, I will check it out.

virtmax:
Are there any plans to eliminate this bug?

jschueller:
Not that I know.

jschueller:
Found it at line 1428 of cobyla.cn both sp and temp are 0.
```
1428          alpha = sp / temp;
```

Add this in your main and run gdb:
```
#include <fenv.h>

int main ()
feenableexcept(FE_INVALID | FE_OVERFLOW);
```

aphirst:
Out of curiosity, is this division by zero also expected to cause issues if solving the same system reimplemented in a different language? (If I have time this evening, I'll attempt a rewrite in Fortran.)

PR #316, opened 2020-03-03
check if divide-by-zero happens

mahyaret:
Thanks for your great repo,
I have noticed that COBYLA algorithm is prone to have divisions by zero. This issue is mentioned here: [https://github.com/stevengj/nlopt/issues/118](https://github.com/stevengj/nlopt/issues/118)

 It actually caused problem for me and I had to add checks for it. This pull request include these checks.

Thanks,
Mahyar


mahyaret:
This is strange since this tests set is passing for me!

PR #323, opened 2020-04-12
Allow force_stop with dimension elimination

### LINKED ISSUE #317: C++ exceptions are ignored when eliminating dimensions
Hi,

When the objective function throws an exception such as `nlopt::forced_stop`, optimization should be gracefully halted and an exception is thrown to the caller. However when bounds coincide for one or more variables in combination with a gradient-free algorithm, this exception is never rethrown.

When stepping through the code, it seems the `force_stop` flag is set in the original `nlopt::opt`, while the check is done for the elimdim wrapper, could this be the issue?

MWE:

```
double myvfunc(const std::vector<double> &x, std::vector<double> &grad,
               void *my_func_data) {
  throw nlopt::forced_stop();
  return x[0];
}

int main() {
  nlopt::opt opt("GN_CRS2_LM", 2);
  std::vector<double> lb{-10.0, 0};
  opt.set_lower_bounds(lb);
  std::vector<double> ub{10, 1};
  opt.set_upper_bounds(ub);
  opt.set_min_objective(myvfunc, NULL);
  opt.set_maxeval(1000);
  opt.set_xtol_rel(1e-4);
  std::vector<double> x{1.234, 0};
  double minf;
// This works as it should, catching and rethrowing the forced_stop exception
  try {
    opt.optimize(x, minf);
    std::cout << "found minimum at f(" << x[0] << "," << x[1]
              << ") = " << std::setprecision(10) << minf << std::endl;
  } catch (std::exception &e) {
    std::cout << "nlopt failed: " << e.what() << std::endl;
  }
  opt.set_upper_bounds({10, 0});
// This catches but never rethrows the exceptions and continues until maxeval is reached.
  try {
    opt.optimize(x, minf);
    std::cout << "found minimum at f(" << x[0] << "," << x[1]
              << ") = " << std::setprecision(10) << minf << std::endl;
  } catch (std::exception &e) {
    std::cout << "nlopt failed: " << e.what() << std::endl;
  }
}
```
This returns 
```
nlopt failed: nlopt forced stop
found minimum at f(1.234,0) = inf
```

jschueller:
Closes #317

### DIFF
--- a/src/api/optimize.c
+++ b/src/api/optimize.c
@@ -851,6 +851,7 @@ nlopt_result NLOPT_STDCALL nlopt_optimize(nlopt_opt opt, double *x, double *opt_
                 goto done;
             }
             elimdim_shrink(opt->n, x, opt->lb, opt->ub);
+            opt->force_stop_child = elim_opt;
         }
 
         ret = nlopt_optimize_(elim_opt, x, opt_f);
@@ -858,6 +859,7 @@ nlopt_result NLOPT_STDCALL nlopt_optimize(nlopt_opt opt, double *x, double *opt_
         if (elim_opt != opt) {
             elimdim_destroy(elim_opt);
             elimdim_expand(opt->n, x, opt->lb, opt->ub);
+            opt->force_stop_child = NULL;
         }
     }
 


jschueller:
hi @stevengj, maybe we could do a patch release ?

PR #324, opened 2020-04-15
Prepare release 2.6.2

jschueller:
@stevengj what about a new patch release ?


stevengj:
Sounds good to me.

stevengj:
tagged

PR #354, opened 2020-08-28
test: Check HAVE_GETOPT_H

jschueller:


stevengj:
I'm worried about the case where the header file `getopt.h` is missing but we still have `getopt` in libc.

Probably we should also rename our`getopt` implementation to `nlopt_getopt` and add `#define getopt nlopt_getopt` etc. to `nlopt-getopt.h`

jschueller:
or check that getopt function is available throught the compiler: `check_function_exists(getopt HAVE_GETOPT)`

stevengj:
If the function is available through the compiler (i.e. linking succeeds), but `getopt.h` is not available, then I think we probably still want to use our own implementation — `testopt.c` will be using our own header file `nlopt-getopt.h` in any case, and we would be implicitly assuming that our header file matches the signature of the `getopt` implementation that the compiler is linking to.

jschueller:
ok

PR #358, opened 2020-10-05
Fix wrong internal link

mzaffalon:


jschueller (changes requested):
it seems to be dashes instead of underscores:
https://nlopt.readthedocs.io/en/latest/NLopt_Algorithms/#augmented-lagrangian-algorithm

mzaffalon:
Thanks for the feedback.
All markdown links seem to be broken: it should e.g. be `NLopt_Reference.md#stochastic-population` instead of `NLopt_Reference#Stochastic_population.md`. It will take me a day to fix them all.

PR #365, opened 2020-11-18
add generic algorithm parameters API

### LINKED ISSUE #329: Max inner iterations for MMA
It would be nice to pass a maximum number of "inner" iterations for the MMA algorithm.

This is especially useful for topology optimization problems.

stevengj:
This PR adds a new API `nlopt_set_param(opt, "name", val)` to set arbitrary named parameters for an optimizer.   This gives us a generic API to set internal parameters of algorithms.

For example, it fixes #329 by adding the API:
```c
nlopt_set_param(opt, "inner_maxeval", 10);
nlopt_set_param(opt, "dual_ftol_rel", 1e-8);
```
to set these parameters for the MMA or CCSA algorithms.

To do:
- [x] Documentation

No Matlab or Fortran bindings for this API.  (Passing strings between Fortran and C is hard to do portably from the C side.)


stevengj:
(Note: append [`?w=1` to the diff URL](https://github.com/stevengj/nlopt/pull/365/files?w=1) to ignore whitespace changes.

PR #385, opened 2021-03-28
Add support for octave-6.2.0

octave-user:
This branch fixes an issue with anonymous functions in GNU Octave 6.2.0. It should be backward compatible with Octave 5.2.0.


pierre-haessig:
Hello,
I'd like to know if this PR is included in the latest release [v2.7.1](https://github.com/stevengj/nlopt/releases/tag/v2.7.1). Since it was released in Dec 2021, I guess it could be, but I don't know how to be sure.

PR #406, opened 2021-07-28
Compatible Apple M1

charlerive:
compile on Apple Silicon


jschueller:
I dont think this should be part of the infrastructure: I mean you can still pass it externally like other compilation or cross-compilation flags.
-1

PR #409, opened 2021-08-11
Fix undefined behaviour in nlopt_result_from_string

### LINKED ISSUE #408: Undefined behaviour for nlopt_result_from_string
I'm observing a segmentation fault with [nlopt 2.7.0 on Arch Linux](https://archlinux.org/packages/community/x86_64/nlopt/) using the following minimal example:

```c
#include "nlopt.h"

#include <string.h>
#include <stdio.h>

int
main (void)
{
    nlopt_result res = nlopt_result_from_string("FAILURE");
    printf("%d\n", res);

    return 0;
}
```

The following commands allow to reproduce the behaviour:

```
❯ gcc --version
gcc (GCC) 11.1.0
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

❯ pkg-config nlopt --modversion
2.7.0
❯ gcc mwe.c -g $(pkg-config nlopt --cflags --libs) && gdb ./a.out
GNU gdb (GDB) 10.2
Copyright (C) 2021 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-pc-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from ./a.out...
(gdb) run
Starting program: /home/awvwgk/projects/src/git/nlopt/a.out 

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7e77a12 in __strcmp_avx2 () from /usr/lib/libc.so.6
(gdb) bt
#0  0x00007ffff7e77a12 in __strcmp_avx2 () from /usr/lib/libc.so.6
#1  0x00007ffff7f49edd in nlopt_result_from_string () from /usr/lib/libnlopt.so.0
#2  0x0000555555555160 in main () at mwe.c:9
```

My guess is that the mapping of the string as implemented here:

https://github.com/stevengj/nlopt/blob/304137ff414be4c0f5abf72cda25bfb3ec2775d1/src/api/general.c#L199-L203

Will go from `0` to `NLOPT_NUM_RESULTS` and fall through the switch case as no entry in the enumerator matches `0` in the first iteration and therefore return `NULL` here:

https://github.com/stevengj/nlopt/blob/304137ff414be4c0f5abf72cda25bfb3ec2775d1/src/api/general.c#L190

Which in turn is passed to `strcmp` and results in [undefined behaviour](https://stackoverflow.com/questions/21865041/in-c-what-exactly-happens-when-you-pass-a-null-pointer-to-strcmp).

awvwgk:
Closes #408

### DIFF
--- a/src/api/general.c
+++ b/src/api/general.c
@@ -185,6 +185,7 @@ const char *nlopt_result_to_string(nlopt_result result)
     case NLOPT_XTOL_REACHED: return "XTOL_REACHED";
     case NLOPT_MAXEVAL_REACHED: return "MAXEVAL_REACHED";
     case NLOPT_MAXTIME_REACHED: return "MAXTIME_REACHED";
+    case NLOPT_NUM_FAILURES:
     case NLOPT_NUM_RESULTS: return NULL;
   }
   return NULL;
@@ -196,7 +197,14 @@ nlopt_result nlopt_result_from_string(const char * name)
   int i;
   if (name == NULL)
     return -1;
-  for (i = 0; i < NLOPT_NUM_RESULTS; ++i)
+  /* Start checking all the failures which are enumerated with negative indices */
+  for (i = -1; i > NLOPT_NUM_FAILURES; --i)
+  {
+    if (strcmp(name, nlopt_result_to_string(i)) == 0)
+      return i;
+  }
+  /* Continue checking all possible successes which are enumerated with positive indices */
+  for (i = 1; i < NLOPT_NUM_RESULTS; ++i)
   {
     if (strcmp(name, nlopt_result_to_string(i)) == 0)
       return i;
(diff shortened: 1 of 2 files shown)


stevengj:
CI failure on Linux and Appveyor are unrelated.

Issue #413, opened 2021-09-14
Suspicious use of long double in slsqp.c


AaronWebster:
I want to get rid of the `long double` type in `slsqp.c` for compatibility with ARM.  If its use is warranted, I can replace it with a Kahan sum.  Otherwise, I'd demote it to a `double`.

stevengj:
`double` should be fine.

AaronWebster:
https://github.com/stevengj/nlopt/pull/414

stevengj:
Closed by #414.

PR #414, opened 2021-09-14
Remove use of long double in slsqp.c

AaronWebster:
Improves compatibility with ARM.  See [https://github.com/stevengj/nlopt/issues/413](https://github.com/stevengj/nlopt/issues/413)

### DIFF
--- a/src/algs/ags/ags.cc
+++ b/src/algs/ags/ags.cc
@@ -53,7 +53,7 @@ int ags_minimize(unsigned n, nlopt_func func, void *data, unsigned m, nlopt_cons
   params.evolventDensity = evolvent_density;
   params.epsR = eps_res;
   params.stopVal = stop->minf_max;
-  params.refineSolution = (bool)ags_refine_loc;
+  params.refineSolution = static_cast<bool>(ags_refine_loc);
 
   ags::NLPSolver solver;
   solver.SetParameters(params);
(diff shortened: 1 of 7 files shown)


jschueller (approved):

stevengj:
On ARM, isn't `long double` simply a synonym for `double`?  I'm not opposed to this change, but I'm confused about why it would improve matters on ARM.

ooosssososos:
Long doubles are just doubles in arm. But on x86 they are sometimes 80 bit. The discrepancy in numerical precision can be a problem when expecting the same output on x86 and ARM, especially since the small error caused by fp differences initially seems to increase over time and can result in sizeable differences in output across platforms. This can cause tests that expect some output to fail

AaronWebster:
From what I understand as it's confined to the FPU on x86 systems long doubles are 80 bit and on arm they are 64 bit; he downcast to double can result in discrepancies which causes solution to diverge.  

https://developer.arm.com/documentation/dui0491/i/C-and-C---Implementation-Details/Basic-data-types

stevengj:
> the downcast to double can result in discrepancies which causes solution to diverge.

Even if we use `double` on all platforms, we cannot guarantee exactly the same results on all platforms — e.g. differences in compiler optimizations (such as whether fused multiply adds are used) may change roundoff errors when you go from one system to the next.

PR #418, opened 2021-09-24
CI: Drop appveyor too

jschueller:


jschueller:
@stevengj could you log into appveyor and delete nlopt so it stops triggering builds ? I switched everything to github actions

PR #457, opened 2022-06-23
Implement C++ style functors as targets for objectives

### LINKED ISSUE #219: C++11 idiom
I found nlopt a great library, but using it through the C++ is really frustrating.
You need to provide a `void*` for passing data to the objective/constraint functions, with all the problemas that may have. 

Also, you need to pass a function pointer, so you cant use lambda functions with captures (that would avoid passing the void*).

Ive written a thin wrapper on top of NLopt which tries to offer a more modern c++ API, enabling the use of lambdas and hiding from the API the use of void*. 

Would you be interested in this to be merge?

dburov190:
This PR implements a wrapper `nlopt::functor_wrapper` for C++ style functors via `std::function`, and two new overloads of `nlopt::set_min_objective`, `nlopt::set_max_objective`.

In order to allow that, a new member field in the `myfunc_data` struct is added: `functor_type functor;`, where `functor_type` is defined as
```c++
typedef std::function<double(unsigned, const double*, double*)> functor_type;
```

This is not introduced as a pointer (like the other function-pointers are) because `std::function` is already a container that stores a pointer, and abstracts it away.

**Important:** note that the signature for the functor does not include `void* data` unlike all other function-pointers. That is because it is assumed that the functor already has all the data it needs.

This PR allows now to write the following:
```c++
class UserDefinedObjective {
  private:
    ImportantData data;
  public:
    UserDefinedObjective() = delete;
    UserDefinedObjective(ImportantData data) :
      data(std::move(data)) {}
    double operator()(unsigned n, const double* x, double* grad) const
    {
      // compute objective(x) and ∇objective(x) using this->data
    }
};

int main()
{
  ImportantData data;
  UserDefinedObjective objective(std::move(data));

  nlopt::opt optimizer;
  // other nlopt settings
  optimizer.set_max_objective(std::move(objective));

  optimizer.optimize(...);

  return 0;
}
```
Same with C++ lambdas, regular functions and even class member functions (check out [std::function](https://en.cppreference.com/w/cpp/utility/functional/function)).

This PR also introduces a CMake macro `NLOPT_add_cpp_test` to quickly add cpp tests, and creates a test `cpp_functor.cxx` to actually test the new functionality.

Closes #219 .

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -158,12 +158,11 @@ endif ()
 if (NLOPT_CXX OR NLOPT_PYTHON OR NLOPT_GUILE OR NLOPT_OCTAVE)
   check_cxx_symbol_exists (__cplusplus ciso646 SYSTEM_HAS_CXX)
   if (SYSTEM_HAS_CXX)
-    check_cxx_compiler_flag ("-std=c++11" SUPPORTS_STDCXX11)
-    if (SUPPORTS_STDCXX11)
-      set (CMAKE_CXX_FLAGS "-std=c++11 ${CMAKE_CXX_FLAGS}")
-      if (NLOPT_CXX)
-        set (NLOPT_CXX11 ON)
-      endif ()
+    set (CMAKE_CXX_STANDARD 11) # set the standard to C++11 but do not require it
+
+    if (NLOPT_CXX)
+      set (CMAKE_CXX_STANDARD_REQUIRED ON) # if we build C++ API, we do need C++11
+      set (NLOPT_CXX11 ON)
     endif ()
   else()
     message (FATAL_ERROR "The compiler doesn't support CXX.")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -162,7 +162,6 @@ if (NLOPT_CXX OR NLOPT_PYTHON OR NLOPT_GUILE OR NLOPT_OCTAVE)
 
     if (NLOPT_CXX)
       set (CMAKE_CXX_STANDARD_REQUIRED ON) # if we build C++ API, we do need C++11
-      set (NLOPT_CXX11 ON)
     endif ()
   else()
     message (FATAL_ERROR "The compiler doesn't support CXX.")
@@ -220,12 +219,27 @@ set (NLOPT_SOURCES
 )
 
 if (NLOPT_CXX)
-  list (APPEND NLOPT_SOURCES src/algs/stogo/global.cc src/algs/stogo/linalg.cc src/algs/stogo/local.cc src/algs/stogo/stogo.cc src/algs/stogo/tools.cc
-        src/algs/stogo/global.h src/algs/stogo/linalg.h src/algs/stogo/local.h src/algs/stogo/stogo_config.h src/algs/stogo/stogo.h src/algs/stogo/tools.h)
-endif ()
-if (NLOPT_CXX11)
-  list (APPEND NLOPT_SOURCES src/algs/ags/data_types.hpp src/algs/ags/evolvent.hpp src/algs/ags/evolvent.cc src/algs/ags/solver.hpp src/algs/ags/solver.cc
-  src/algs/ags/local_optimizer.hpp src/algs/ags/local_optimizer.cc src/algs/ags/ags.h src/algs/ags/ags.cc)
+  list (APPEND NLOPT_SOURCES
+    src/algs/stogo/global.cc
+    src/algs/stogo/linalg.cc
+    src/algs/stogo/local.cc
+    src/algs/stogo/stogo.cc
+    src/algs/stogo/tools.cc
+    src/algs/stogo/global.h
+    src/algs/stogo/linalg.h
+    src/algs/stogo/local.h
+    src/algs/stogo/stogo_config.h
+    src/algs/stogo/stogo.h
+    src/algs/stogo/tools.h
+    src/algs/ags/data_types.hpp
+    src/algs/ags/evolvent.hpp
+    src/algs/ags/evolvent.cc
+    src/algs/ags/solver.hpp
+    src/algs/ags/solver.cc
+    src/algs/ags/local_optimizer.hpp
+    src/algs/ags/local_optimizer.cc
+    src/algs/ags/ags.h
+    src/algs/ags/ags.cc)
 endif ()
 
 install (FILES ${NLOPT_HEADERS} DESTINATION ${RELATIVE_INSTALL_INCLUDE_DIR})
--- a/src/api/general.c
+++ b/src/api/general.c
@@ -46,9 +46,11 @@ static const char nlopt_algorithm_names[NLOPT_NUM_ALGORITHMS][256] = {
 #ifdef NLOPT_CXX
     "StoGO (global, derivative-based)",
     "StoGO with randomized search (global, derivative-based)",
+    "AGS (global, no-derivative)"
 #else
     "StoGO (NOT COMPILED)",
     "StoGO randomized (NOT COMPILED)",
+    "AGS (NOT COMPILED)"
 #endif
     "original L-BFGS code by Nocedal et al. (NOT COMPILED)",
     "Limited-memory BFGS (L-BFGS) (local, derivative-based)",
@@ -83,11 +85,6 @@ static const char nlopt_algorithm_names[NLOPT_NUM_ALGORITHMS][256] = {
     "Sequential Quadratic Programming (SQP) (local, derivative)",
     "CCSA (Conservative Convex Separable Approximations) with simple quadratic approximations (local, derivative)",
     "ESCH evolutionary strategy",
-#ifdef NLOPT_CXX11
-    "AGS (global, no-derivative)"
-#else
-    "AGS (NOT COMPILED)"
-#endif
 };
 
 const char *NLOPT_STDCALL nlopt_algorithm_name(nlopt_algorithm a)
--- a/src/api/nlopt-in.hpp
+++ b/src/api/nlopt-in.hpp
@@ -103,6 +103,24 @@ namespace nlopt {
       nlopt_munge munge_destroy, munge_copy; // non-NULL for SWIG wrappers
     } myfunc_data;
 
+    void* alloc_myfunc_data_with_nulls() {
+      // need to return void* otherwise SWIG doesn't compile because
+      // myfunc_data is private
+      myfunc_data *d = new myfunc_data;
+      if (!d) throw std::bad_alloc();
+
+      d->o             = this;
+      d->f             = NULL;
+      d->f_data        = NULL;
+      d->mf            = NULL;
+      d->vf            = NULL;
+      d->munge_destroy = NULL;
+      d->munge_copy    = NULL;
+
+      return reinterpret_cast<void*>(d);
+    }
+
+
     // free/destroy f_data in nlopt_destroy and nlopt_copy, respectively
     static void *free_myfunc_data(void *p) {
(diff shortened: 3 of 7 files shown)


dburov190:
I'd add, I didn't implement any tests because I didn't find a consistent style, there's one file named `t_tutorial.cxx` -- should I add to that file? or create a new one?

Anyway, however I tested this in my own code: it compiles and runs OK.

jschueller:
I'd say create another test: t_tutorial.cxx aims at reproducing the example from the documentation.

You will need to create a cmake macro to reuse these lines for another test:
https://github.com/stevengj/nlopt/blob/master/test/CMakeLists.txt#L3-L10

dburov190:
@jschueller Done. Wrote a macro and a (somewhat) comprehensive test for the new syntax.

dburov190:
I looked at the failing log output again, and I can't get it why it failed. It says that `t_tutorial.exe` and `cpp_functor.exe` were not found, but `testopt.exe` was also not found, and I didn't change anything about that test. It also complained about private struct `myfunc_data`, but  again, I didn't change anything except adding a new field. It would be hard to debug this for me, unfortunately, as I don't own a Windows machine (fortunately!).

dburov190:
I suspect it was because I wrote a function `alloc_myfunc_data_with_nulls()` that returned a `myfunc_data*` pointer. I changed it to `void*`. Hopefully SWIG compiles now. thanks!

src/api/nlopt-in.hpp
```diff
@@ -89,10 +98,29 @@ namespace nlopt {
...
+      myfunc_data *d = new myfunc_data;
+      if (!d) throw std::bad_alloc();
+
+      d->o             = this;
+      d->f             = NULL;
+      d->f_data        = NULL;
+      d->mf            = NULL;
+      d->vf            = NULL;
+      d->munge_destroy = NULL;
+      d->munge_copy    = NULL;
```
stevengj, line 118: Can't you just use `new myfunc_data()` (note [parentheses](https://stackoverflow.com/questions/620137/do-the-parentheses-after-the-type-name-make-a-difference-with-new)) to [automatically zero-initialize](https://en.cppreference.com/w/cpp/language/zero_initialization)?

src/api/nlopt-in.hpp
```diff
@@ -89,10 +98,29 @@ namespace nlopt {
     typedef struct {
       opt *o;
       mfunc mf; func f; void *f_data;
+      functor_type functor;
```
stevengj, line 101: It seems like we'll need to make sure that this is initialized in the other places that call `new myfunc_data`, since otherwise `delete` on the data will call the destructor of an uninitialized value, no?

src/api/nlopt-in.hpp
```diff
@@ -34,6 +34,8 @@
 #include <cstdlib>
 #include <cstring>
 #include <cmath>
+#include <utility> // std::move
+#include <functional> // std::function
```
stevengj, line 38: Note that the [current build scripts don't require C++11](https://github.com/stevengj/nlopt/blob/17826b8dd812c1b8cc10042432e60d06ea35e82d/CMakeLists.txt#L162-L167), which is needed for `std::function`.

We'll either need to change the build to require C++11, or add some magic to disable the `std::function` stuff when `nlopt.hpp` is generated from `nlopt-in.hpp` if C++11 is not found.

dburov190: True, I overlooked this. My feeling would be to require C++11 (after all, we're in 2022, eleven (!) years after it was released), but since this is your repository, I defer this decision to you (: If you say drop `std::function`, I could rewrite the functor functionality using templates I believe.

src/api/nlopt-in.hpp
dburov190: This is actually OK because `std::function` has a default constructor (which simply stores a null pointer), so even when `new myfunc_data` is called, it is properly constructed; and then it's safe to delete it.

src/api/nlopt-in.hpp
dburov190: @jschueller @stevengj 
OK, `CMakeLists.txt` script now requires C++11 for C++11 API.

dburov190:
bump
Anything else left to do?

dburov190:
@jschueller @stevengj 

Could you guys merge this in please? if everything looks OK to you

CMakeLists.txt
```diff
@@ -158,12 +158,11 @@ endif ()
...
-    if (SUPPORTS_STDCXX11)
-      set (CMAKE_CXX_FLAGS "-std=c++11 ${CMAKE_CXX_FLAGS}")
-      if (NLOPT_CXX)
-        set (NLOPT_CXX11 ON)
-      endif ()
+    set (CMAKE_CXX_STANDARD 11) # set the standard to C++11 but do not require it
+
+    if (NLOPT_CXX)
+      set (CMAKE_CXX_STANDARD_REQUIRED ON) # if we build C++ API, we do need C++11
+      set (NLOPT_CXX11 ON)
```
jschueller, line 165: NLOPT_CXX11 can be dropped everywhere now as implied by NLOPT_CXX

dburov190: fixed, thnx

dburov190:
bumping it again

src/api/nlopt-in.hpp
```diff
@@ -299,6 +343,14 @@ namespace nlopt {
...
+    void set_min_objective(functor_type functor) {
+      myfunc_data *d =
+        reinterpret_cast<myfunc_data*>(alloc_myfunc_data_with_nulls());
+
+      d->functor = std::move(functor);
+      mythrow(nlopt_set_min_objective(o, functor_wrapper, d)); // d freed via o
+    }
+
     void set_max_objective(func f, void *f_data) {
       myfunc_data *d = new myfunc_data;
```
stevengj, line 355: Shouldn't this (and similar cases) now be `myfunc_data *d = new myfunc_data()` to zero-initialize now, to be sure that the `functor` field gets a default initializer?

dburov190: Not if you zero-initialize manually the rest of the fields (which are raw pointers): since `std::function` is an STL container, it has a default constructor, which is called when `myfunc_data` is default-initialized (or zero-initialized), so `functor` will never be in an undefined state.

But what I could do instead, is rewrite those methods with `alloc_myfunc_data_with_nulls` and not deal with `new` explicitly in any of them.

dburov190: okay, done! (:

src/api/nlopt-in.hpp
```diff
@@ -89,10 +98,29 @@ namespace nlopt {
...
       opt *o;
       mfunc mf; func f; void *f_data;
+      functor_type functor;
       vfunc vf;
       nlopt_munge munge_destroy, munge_copy; // non-NULL for SWIG wrappers
     } myfunc_data;
 
+    static void* alloc_myfunc_data_with_nulls() {
+      // need to return void* otherwise SWIG doesn't compile because
+      // myfunc_data is private
```
stevengj, line 108: This is a private method, no?  So casts shouldn't be needed for SWIG.

dburov190: It is private right now, but I think you might be right, in the sense it wouldn't compile when it was public and then I both changed it to private and to `void*` returning.
Unfortunately, I have no way of testing this locally, could you run the CI please?

dburov190:
Hey, could you re-run the CI please?

dburov190:
OK, does everything look good now?

dburov190:
Hello?.. 🥲

jschueller (approved):

dburov190:
hi! any progress on this?

test/cpp_functor.cxx
```diff
@@ -0,0 +1,209 @@
...
+  LinearRegression objective_1(form);
+  SineRegression   objective_2(std::move(form));
+
+  nlopt::opt optimizer("LD_MMA", dimension);
+  optimizer.set_xtol_rel(1e-4);
+  optimizer.set_maxeval(1000);
+
+
+  // linear regression optimization
+  optimizer.set_min_objective(std::move(objective_1));
```
stevengj, line 169: My understanding is that `std::move` leaves the source object (`objective_1`) in a "valid but undefined state", which seems like not the sort of thing one would always want to do — what if the caller wants to use the objective function for something other than optimization?

In general, we should be clear about the resource management here.  Who is responsible for freeing the resources associated with the objective functor?  Do they get freed when the functor leaves the scope where it was defined, or do they get freed when the `optimizer` object is destroyed?

dburov190: In this example, I simply used `std::move` just because I myself don't need the `objective_1` afterwards in the `main` function, so it's better to move it rather than copy. But in practice, users should decide for themselves what they want: to make a copy or move the object. Since `set_min_objective` accepts the parameter by value, it will make a copy of the argument unless it's passed using `std::move` (or is an otherwise RHS expression, like nameless lambda). OK, this is a slight simplification because in reality `std::function`'s constructor is called in between, but the copy-or-move logic is still intact -- because `std::function` has a move ctor.

In general, since `set_min_objective`, `set_max_objective` accept a functor by value, it is assumed that `optimizer` is responsible for memory management of its own copy of the functor, but the outer one should be managed by the user. I believe this is fairly common, and actually the `std::move` allows the user to decide whether s/he wants to keep the object or pass it altogether, instead of hardcoding this into a library. Again, in this example I chose to "move" the object because I just don't need it afterwards. But we can also write simply
```cpp
optimizer.set_min_objective(objective_1);
```
and then the `objective_1` is usable (and a copy is made).

As for the inner workings, since functor is an STL-container, functor will get destroyed when `std::function`'s destructor is called. Which, in turn, happens when `myfunc_data` is freed. And that, in turn, is already well-designed within the library.

Do you agree with this approach?

stevengj: That seems fine to me.  We shouldn't use `std::move` in the documentation examples, however, because people are likely to copy that blindly and will then get into trouble.

dburov190: oh, okay! I could change that if you want? in a new PR? and thank you for merging!

stevengj: A new PR to clarify this would be fine.

src/api/general.c
```diff
@@ -83,11 +85,6 @@ static const char nlopt_algorithm_names[NLOPT_NUM_ALGORITHMS][256] = {
     "Sequential Quadratic Programming (SQP) (local, derivative)",
     "CCSA (Conservative Convex Separable Approximations) with simple quadratic approximations (local, derivative)",
     "ESCH evolutionary strategy",
-#ifdef NLOPT_CXX11
-    "AGS (global, no-derivative)"
-#else
-    "AGS (NOT COMPILED)"
-#endif
```
NogginBops, line 90: This PR broke `nlopt_algorithm_name` as it relied on this ordering that was changed...

src/api/general.c
```diff
@@ -46,9 +46,11 @@ static const char nlopt_algorithm_names[NLOPT_NUM_ALGORITHMS][256] = {
 #ifdef NLOPT_CXX
     "StoGO (global, derivative-based)",
     "StoGO with randomized search (global, derivative-based)",
+    "AGS (global, no-derivative)"
 #else
     "StoGO (NOT COMPILED)",
     "StoGO randomized (NOT COMPILED)",
+    "AGS (NOT COMPILED)"
```
NogginBops, line 53: The missing `,` at the end of these items cause them to be merged with the item after which causes there to be one too few items in this array.

PR #469, opened 2022-08-22
Casts are added. Some compiler warnings (e.g. 'C4267', 'C4244') of MS…

knut-o:
Some compiler warnings with MS Visual Studio compiler are avoided.


jschueller:
hello, could you rebase to master ?

PR #497, opened 2023-02-03
Fix/cmake guile

pavanakumar:
FindGuile.cmake does not seem to work everywhere. The pkg-config approach to guile is quite robust and is available in most platforms. This is a cleaned up CMake script for NLopt Guile wrapper.

### DIFF
--- a/cmake/FindGuile.cmake
+++ b/cmake/FindGuile.cmake
@@ -12,87 +12,34 @@
 # GUILE_SITE_DIR         - site dir
 # GUILE_EXTENSION_DIR    - extension dir
 # GUILE_ROOT_DIR         - prefix dir
+find_package(PkgConfig REQUIRED)
 
-# Look for the header file
-# Look for guile-2.2 first, then 2.0, then 1.8
-# Macports for OSX puts things in /opt/local
-find_path (GUILE_INCLUDE_DIR libguile.h
-  PATH_SUFFIXES
-    guile/3.0
-    guile/2.2
-    guile/2.0
-    guile/1.8
-    libguile
-    guile
-  HINTS /opt/local/include
-)
-
-# Look for the library
-find_library (GUILE_LIBRARY NAMES guile-3.0 guile-2.2 guile-2.0 guile
-  HINTS
-    /opt/local/lib
-)
-
-
-set (GUILE_LIBRARIES ${GUILE_LIBRARY})
-set (GUILE_INCLUDE_DIRS ${GUILE_INCLUDE_DIR})
-
-
-# check guile's version if we're using cmake >= 2.6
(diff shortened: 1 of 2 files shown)


jschueller:
on what platform does it not work ?
maybe we dont watn to force require pkgconfig
anyways this should stay in findguile

pavanakumar:
> on what platform does it not work ? 

This is failing on my Mac 11.7 (BigSur) and I installed guile using homebrew

> maybe we dont watn to force require pkgconfig anyways this should stay in findguile

I do not agree with requiring pkg-config because if you look at the FindGuile.cmake,

```
if (GUILE_CONFIG_EXECUTABLE)
  execute_process (COMMAND ${GUILE_CONFIG_EXECUTABLE} info prefix
                    OUTPUT_VARIABLE GUILE_ROOT_DIR
                    OUTPUT_STRIP_TRAILING_WHITESPACE)
...
```
It is using the `GUILE_CONFIG_EXECUTABLE=guile-config` command that in turn is a wrapper over `pkg-config`. On certain platforms the `execute_command` puts extra quotes around `OUTPUT_VARIABLE` and that is still an unresolved issue with CMake in my experience. Directly using `pkg-config` from the CMake packages is much safer and is better supported within CMake.

I do agree that this should be inside FindGuile.cmake. Will make these changes and squash it into a single commit and update the pull request.

pavanakumar:
> > It is using the GUILE_CONFIG_EXECUTABLE=guile-config command that in turn is a wrapper over pkg-config.
> 
> No, `guile-config` is _not_ a wrapper around `pkg-config`. It is installed as a part of guile itself (e.g. if you build Guile from source and do `make install`, and in fact is [written in Scheme](https://github.com/cky/guile/blob/89ce9fb31b00f1f243fe6f2450db50372cc0b86d/meta/guile-config.in). It's the documented "official" way to find out how to link against guile.

It does look like a wrapper over the pkg-config on my platform (written in scheme).

```
prompt> guile-config info --cflags 
error: (“/opt/homebrew/opt/pkg-config/bin/pkg-config” “--variable=--cflags” “guile-3.0”) exited with non-zero error code 127
```

And please read line 28 of the `guile-config.in` you sent me it reads as follows

```
;;; This script has been deprecated. Just use pkg-config.
```
> 
> Using `guile-config` will fail if you haven't installed it into your `PATH`, but then you should either set your `PATH` appropriately or pass `GUILE_CONFIG_EXECUTABLE=/path/to/guile-config` to `cmake`.

This doesn't help

jschueller:
hello @pavanakumar, could you retry with master ? I got it running with guile from homebrew

jschueller:
See if it works with 9f8ddbf1886ec0838bf04b6a3e290d030fe9e244

PR #508, opened 2023-05-02
MLSL should print at least a little information

TiborGY:
As of now, MLSL does not print anything that would inform the user of its progress. This is irritating as such an optimization may easily be running for days without any output at all.

This PR adds some fairly minimal text output to MLSL, to let the user know that there is no problem, MLSL is running OK.

### DIFF
--- a/src/algs/mlsl/mlsl.c
+++ b/src/algs/mlsl/mlsl.c
@@ -64,6 +64,7 @@
 #include <stdlib.h>
 #include <math.h>
 #include <string.h>
+#include <stdio.h>
 
 #include "redblack.h"
 #include "mlsl.h"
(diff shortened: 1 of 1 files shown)


jschueller:
libs should not print anything to stdout/stderr by default

TiborGY:
> libs should not print anything to stdout/stderr by default

That is a matter of taste, but when I call a function that might not return for a few days I do prefer at least a little output.

jschueller:
it should be optional, with a callback mechanism or something

stevengj:
You can print anything you want from your objective function.  e.g. I typically add a few lines to my objective function to print a status output, e.g. maybe the value of the objective function and the best value of the objective function so far, maybe an iteration count.

However, I agree that it is sometimes useful to print some algorithm-specific status status updates.  The way to do this is

1. Add an `int verbosity` parameter  to `mlsl_minimize`.
2. Only print when `verbosity > 0`.  Maybe print even more output if `verbosity > 1`.  e.g. see how MMA does it: https://github.com/stevengj/nlopt/blob/a36a0ad95dc2120bd587a3e2ab71c97319d1ff5c/src/algs/mma/mma.c#L284-L290
3. When `mlsl_minimize` is called, pass  `(int) nlopt_get_param(opt, "verbosity", 0)` to `mlsl_minimize`.  That will let the user control the verbosity by setting the `verbosity` algorithm-specific parameter. 
 Again, similar to how MMA does it https://github.com/stevengj/nlopt/blob/a36a0ad95dc2120bd587a3e2ab71c97319d1ff5c/src/api/optimize.c#L674

PR #509, opened 2023-05-08
Remove equality constriant number check when adding constraint

Shuenhoy:
Hi,

I have noticed that when the number of equality constraints exceeds the number of variables, NLOPT refuses to add them. However, this does not necessarily mean that there is no solution as there can be redundancy in the constraints that may be non-trivial to remove for the non-linear situation from the user side. Moreover, even if the number of equality constraints is less than the number of variables, this check cannot guarantee the solution's existence.  So this check seems not to make a lot of sense.

Therefore, I removed the check in this PR, or do you prefer a CMake option and preprocess directive that can conditionally disable it?


stevengj:
LGTM.

Issue #368, opened 2020-11-22
nloptr SLSQP returning successful flag despite constraints being unmet


ssmil:
Hello,

I'm having an issue where nlopt is returning a successful exit flag, but only 1 of my 6 constraints are met. I am running many optimizations at once and using the exit flags as automatic checks, so this is causing me some significant issues. I've attached a file with the code I'm running. 

I'm using nloptr version 1.2.2.2, algorithm "NLOPT_LD_SLSQP".

iteration: 16
	x = ( 1.000000, 1.000000, 1.552908, 1.855128 )
	f(x) = -1.907547
	h(x) = ( 15.437643, 5.305731 )
	g(x) = ( -2.609576, 0.328356, 0.500000, 3.871769 )

> rest.CI.GI$status
[1] 3

[nloptr flag issue.zip](https://github.com/stevengj/nlopt/files/5580094/nloptr.flag.issue.zip)

PR #514, opened 2023-06-04
SLSQP search stopped too early on tutorial constraints

edwintorok:
I noticed that the python and C++ tutorials gave different answers when used with SLSQP:
    SLSQP was giving the wrong answer on the tutorial in 't_python.py':
    ```
    optimum at  [0.50777849 4.76252907]
    minimum value =  2.182321944309753
    ```
    instead of:
    ```
    optimum at  [0.33333333 0.29629629]
    minimum value =  0.5443310523133087
    ```

They are implemented slightly differently, using float multiplication vs `** 2` and `** 3`, which result in slightly different floating point values, however that shouldn't cause SLSQP to fail this badly.
The problem isn't with  the Python wrapper, because if I reimplement the same function for 'SciPy' it gives the correct answer.

After printing the individual 'f' evaluations I noticed that it stopped the search due to hitting 'xtol_rel', but the function was infeasible at that point (so it returned the last feasible one, which is way off the optimal value).
If I instead let the search continue until it finds a feasible point again then it finds the correct answer.
  
The problem was hinted to at https://github.com/stevengj/nlopt/issues/368#issuecomment-1196810479, which mentions that perhaps `feasible_cur` should be used, and I think as this example shows that would be the correct approach
(the original SLSQP fortran function also had a check for 'h3', which looks similar to the feasibility check done here).

While fixing the bug I also tweaked the tutorial function to accept the algorithm as parameter and run this during `make test` to check a few algorithms (those that accept inequality constraints and infinite bounds).

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -40,6 +40,13 @@ foreach (algo_index RANGE 43)
         set (enable_ FALSE)
       endif ()
     endif ()
+    # 30 and 31 fail with NLOPT_ROUNDOFF_LIMITED depending on random starting point
+    if (algo_index STREQUAL 30)
+        set (enable_ FALSE)
+    endif ()
+    if (algo_index STREQUAL 31)
+        set (enable_ FALSE)
+    endif ()
     # cxx ags
     if (NOT NLOPT_CXX AND algo_index STREQUAL 43)
       set (enable_ FALSE)
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -29,7 +29,7 @@ if (NLOPT_CXX)
   set_target_properties(testopt PROPERTIES LINKER_LANGUAGE CXX)
 endif ()
 
-foreach (algo_index RANGE 29)# 43
+foreach (algo_index RANGE 43)
   foreach (obj_index RANGE 1)# 21
     set (enable_ TRUE)
     # cxx stogo
@@ -44,6 +44,10 @@ foreach (algo_index RANGE 29)# 43
     endif ()
     # L-BFGS
     if (algo_index STREQUAL 10)
+      set (enable_ FALSE)
+    endif ()
+    if (algo_index STREQUAL 36 OR algo_index STREQUAL 37
+        OR algo_index STREQUAL 38 OR algo_index STREQUAL 39)
       set (enable_ FALSE)
     endif ()
     if (enable_)
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -60,7 +60,7 @@ foreach (algo_index RANGE 43)
       set (enable_ FALSE)
     endif ()
     if (enable_)
-      add_test (NAME testopt_algo${algo_index}_obj${obj_index} COMMAND testopt -r 0 -a ${algo_index} -o ${obj_index})
+      add_test (NAME testopt_algo${algo_index}_obj${obj_index} COMMAND testopt -r 0 -m 1e-3 -a ${algo_index} -o ${obj_index})
       if (CMAKE_HOST_SYSTEM_NAME MATCHES Windows)
         set_tests_properties (testopt_algo${algo_index}_obj${obj_index} PROPERTIES ENVIRONMENT "PATH=${PROJECT_BINARY_DIR}\\${CMAKE_BUILD_TYPE};$ENV{PATH}")  # to load dll
       endif ()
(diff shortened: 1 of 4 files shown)


stevengj:
LGTM.

jschueller:
some tests fail

edwintorok:
Interesting it fails only on windows but not Linux. I'll take a look, but perhaps for now I'll mark those tests as skipped and fix the problems in another PR (those algorithms were not being tested at all previously I think).

edwintorok:
I can reproduce one of the failures on Linux: the starting point is randomly generated and some starting points cause the algorithm to fail with `NLOPT_ROUNDOFF_LIMITED`.
This will eventually fail: `while ./testopt -a 30 -o 1; do echo -n ;done`

stevengj:
> This will eventually fail: `while ./testopt -a 30 -o 1`; do echo -n ;done.

You aren't passing any stopping tolerance to `testopt`?

That being said, the tests should probably pass `-r 0` to use a fixed random seed, in order to lessen the chance of a weird random event biting us on repeated test runs.

test/CMakeLists.txt
```diff
@@ -46,6 +55,10 @@ foreach (algo_index RANGE 29)# 43
     if (algo_index STREQUAL 10)
       set (enable_ FALSE)
     endif ()
+    if (algo_index STREQUAL 36 OR algo_index STREQUAL 37
+        OR algo_index STREQUAL 38 OR algo_index STREQUAL 39)
+      set (enable_ FALSE)
+    endif ()
     if (enable_)
       add_test (NAME testopt_algo${algo_index}_obj${obj_index} COMMAND testopt -r 0 -a ${algo_index} -o ${obj_index})
```
stevengj, line 63:
```suggestion
      add_test (NAME testopt_algo${algo_index}_obj${obj_index} COMMAND testopt -r 0 -m 1e-3 -a ${algo_index} -o ${obj_index})
```
see if we can fix `ROUNDOFF_LIMITED` errors by adding a tolerance here (rather than just running for 1000 iterations).

stevengj:
Still failing on Windows 😢

kkofler:
Sad that this is still stuck after more than a year. SLSQP is unusable with any good stopping criteria (i.e., not just maximum number of iterations or time) due to this bug. The fix is failing tests, but the code in 2.8.0 is clearly incorrect and only passes the tests by accident. (In fact, the claimed SLSQP fix in 2.8.0 basically does not work due to the typo fixed by this PR.)

kkofler:
SLSQP patched with commit c168a7b79ad83b7b970a409500e45eccd1166653 works great for me. SLSQP without the patch just plain does not work if you use any `nlopt_set_[fx]tol_*`.

kkofler:
Looks like disabling algo_index 32 in addition to 30 and 31 should make the tests pass.

kkofler:
Also, the test suite failure is not even a real regression. Tests for algorithm 32 simply did not run before commit a6ceab9c5002521d36ea1836e6871fcb6aeaf6b2, which is also part of this pull request. So I do not see how this is a valid reason for not merging or cherry-picking the bugfix commit c168a7b79ad83b7b970a409500e45eccd1166653.

kkofler:
Checking the enum, algorithms 30 to 33 are the AUGLAG algorithms:
```
    NLOPT_LN_AUGLAG,
    NLOPT_LD_AUGLAG,
    NLOPT_LN_AUGLAG_EQ,
    NLOPT_LD_AUGLAG_EQ,
```
so the test failure is not even **in** SLSQP (the wrapped algorithm is not SLSQP either, but `NLOPT_LN_COBYLA` or `NLOPT_LD_MMA`), so why is it withholding an SLSQP bugfix?

AugLag is known to have issues, see, e.g., #538, but that should be addressed in that issue and not blocking an SLSQP fix.

Commit d4794f412fc8f853d91bff7b713c51a58f1bd9ac disables `NLOPT_LN_AUGLAG` and `NLOPT_LD_AUGLAG` in the test suite, but `NLOPT_LN_AUGLAG_EQ` and `NLOPT_LD_AUGLAG_EQ` are still tested, and `NLOPT_LN_AUGLAG_EQ` is the one that is failing on Windows. It is straightforward to disable 32 and 33 in the tests as well.

It is sad that 2.8.0 shipped without this SLSQP fix, which was submitted more than a year before the 2.8.0 release, all the more because 2.8.0 claims to fix SLSQP, but the fix is ineffective without this followup fix. But we cannot change the past. What we need now is a quick 2.8.1 release with this fix.

Since the original submitter has not done it, do you want me to submit a cleaned-up pull request? a) One with only the SLSQP fix? Or b) one with the additional enabled tests, but all 4 AUGLAG tests (not just 2 of them) disabled until further notice?

kkofler:
Or c) with the SLSQP fix (as in a) + enabling only SLSQP (should be algo_index 40 if I counted correctly) in the testsuite, not any other unrelated algorithms.

Please let me know what your preferences are so I can submit a PR.

jschueller:
I integrated your slsqp fix + the python test update

PR #515, opened 2023-06-10
Some improvements to CMake

besser82:
As described by commit messages.


CMakeLists.txt
```diff
@@ -84,11 +84,16 @@ if (POLICY CMP0086)
   # UseSWIG honors SWIG_MODULE_NAME via -module flag
   cmake_policy(SET CMP0086 NEW)
 endif ()
+if (POLICY CMP0148)
+  # The FindPythonInterp and FindPythonLibs modules are removed
```
jschueller, line 91: are you sure ? we dont use FindPythonInterp anymore

cmake/generate-cpp.cmake
```diff
@@ -1,4 +1,7 @@
-cmake_minimum_required (VERSION 3.2)
+if (POLICY CMP0007)
+  # Do not ignoere empty list items
```
jschueller, line 2: ignoere

CMakeLists.txt
besser82: Yes, for that reason I've explicitly set the policy - that has been introduced with CMake-3.27 - to `NEW` behavior.

cmake/generate-cpp.cmake
besser82: Changed to `list command no longer ignores empty elements` in rebased commit.

CMakeLists.txt
besser82: See [CMake docs for CMP0148](https://cmake.org/cmake/help/latest/policy/CMP0148.html) for further reference.

besser82: > The OLD behavior of this policy is for find_package(PythonInterp) and find_package(PythonLibs) to load the deprecated modules. The NEW behavior is for uses of the modules to fail as if they do not exist.
>
> This policy was introduced in CMake version 3.27. CMake version 3.27.0-rc1 warns when the policy is not set and uses OLD behavior. Use the [cmake_policy()](https://cmake.org/cmake/help/latest/command/cmake_policy.html#command:cmake_policy) command to set it to OLD or NEW explicitly.

besser82:
Thank you!


> I applied most of these, except CMP0148

You forgot the RPATH option.  To cover that, I've opened [another PR](https://github.com/stevengj/nlopt/pull/523).

PR #523, opened 2023-07-19
CMake: Add an option to disable RPATH on user request.

besser82:
Honor -DNLOPT_RPATH=(ON|OFF) to link RPATH into installed NLopt libraries (Default: ON).


jschueller:
you can already set these 2 options since I put them in the cache, so another option on top of that is maybe overkill

jschueller:
hi, I think this should not go in, but it doesnt mean we cannot rediscuss it if you feel I'm wrong

PR #535, opened 2023-10-30
Use the anonymous namespace for a file-local class in stogo

jwnimmer-tri:
BTW I saw that in `src/algs/ags/solver.cc` all of the code inside the anonymous was indented by an extra level.  Let me know if you'd like me to reformat this patch in the same way.


stevengj:
This seems fine.

stevengj:
Thanks!

PR #537, opened 2023-11-21
Incorrect constant name for unscaled variant of DIRECT algorithm

xperroni:
The constant for the unscaled variant of the DIRECT algorithm is incorrectly given as NLOPT_GNL_DIRECT_NOSCAL, when it is in fact NLOPT_GN_DIRECT_NOSCAL.


stevengj (approved):

stevengj:
Thanks!

PR #547, opened 2024-01-23
Exclude nlopt.f from headers when fortran is disabled

abram:
Currently attempting to create an Xcode project with cmake -G Xcode gives the following error:

```
  The custom command generating

    nlopt-build/nlopt.f

  is attached to multiple targets:

    generate-fortran
    nlopt

  but none of these is a common dependency of the other(s).  This is not
  allowed by the Xcode "new build system".
```

This removes nlopt.f from headers unless fortran support is explicitly disabled, so `generate-fortran` won't be considered a dependency of non-fortran builds.


abram:
Hello! I was just doing some code cleanup and realized I'd never submitted a pull request for this small patch I made to the nlopt CMakeLists file a couple of years ago. Thanks for the wonderful project!

jschueller:
7218c80

PR #558, opened 2024-07-08
Fix nlopt_algorithm_name

NogginBops:
Fixes issues introduced in #457

There where missing commas at the end of `"AGS (global, no-derivative)"` causing it to be merged with the next string.
It's ordering was also changed without considering that the order of the items in the array mattered.

This PR fixes the ordering and makes sure that there are terminating commas for each entry.


stevengj:
LGTM, thanks!

PR #566, opened 2024-09-15
Fix small copy paste bug in NLopt_Reference.md

ludwigkr:
The current documentation shows two times the same function. So, I guess this is what it probably meant by the original author.


stevengj:
Thanks!

PR #578, opened 2024-11-29
Add Java binding using the existing SWIG infrastructure

kkofler:
CMakeLists.txt: Add NLOPT_JAVA option. If set, check for C++, JNI, and Java.

src/swig/nlopt-java.i: New SWIG input file (not standalone, but to be included into nlopt.i). Contains the Java-specific declarations.

src/swig/nlopt.i: Add syntax highlighting declaration for KatePart-based editors. Instantiate vector<double> as doublevector instead of nlopt_doublevector for SWIGJAVA, because everything is in an nlopt package in Java, so the nlopt_ part is redundant. Ignore nlopt_get_initial_step because at least for Java, SWIG wants to generate a binding for it, using SWIGTYPE_p_* types that cannot be practically used. (It is in-place just like nlopt::opt::get_initial_step, which is already ignored.) For SWIGJAVA, %include nlopt-java.i.

src/swig/CMakeLists.txt: Set UseSWIG_MODULE_VERSION to 2 so that UseSWIG cleans up old generated source files before running SWIG, useful for Java. (In particular, it prevents obsolete .java files from old .i file versions, which might not even compile anymore, from being included in the Java compilation and the JAR.) If JNI, Java, and SWIG were found, generate the Java binding with SWIG, and compile the JNI library with the C++ compiler and the Java JAR with the Java compiler.

src/swig/glob_java.cmake: New helper script whose only purpose is to invoke the CMake file(GLOB ...) command at the correct stage of the build process, after the SWIG run, not when CMake is run on the main CMakeLists.txt, because the latter happens before anything at all is built. The script is invoked through cmake -P by an add_custom_command in CMakeLists.txt, whose dependencies order it to the correct spot of the build. This is the only portable way to automatically determine which *.java files SWIG has generated from the *.i files. The result is written to java_sources.txt, which is dynamically read by the add_jar command thanks to the @ indirection.

test/t_java.java: New test. Java port of t_tutorial.cxx/t_python.py.

test/CMakeLists.txt: If JNI, Java >= 1.8, and SWIG were found, run the t_java test program with the algorithms 23 (MMA), 24 (COBYLA), 30 (augmented Lagrangian), and 39 (SLSQP). All 4 tests pass. (Java < 1.8 should be supported by the binding itself, but not by the test.)

This code is mostly unrelated to the old unmaintained nlopt4j binding (https://github.com/dibyendumajumdar/nlopt4j), which used handwritten JNI code. Instead, it reuses the existing SWIG binding infrastructure, adding Java as a supported language to that. Only the code in the func_java and mfunc_java wrappers is loosely inspired by the nlopt4j code. The rest of the code in this binding relies mostly on SWIG features and uses very little handwritten JNI code, so nlopt4j was not useful as a reference for it. This binding is also backwards-incompatible with nlopt4j due to differing naming conventions.

Note that this binding maps the C++ class and method names identically to Java. As a result, it does not use idiomatic Java case conventions, but uses snake_case for both class and method names instead of the usual UpperCamelCase for class names and lowerCamelCase for method names in Java. The C++ namespace nlopt is mapped to the Java package nlopt.


kkofler:
Does the binding make sense to you as is? Or do you think I should use the SWIG `%rename` feature aggressively to convert all the class and method names to the Java camel-case conventions?

stevengj:
It would be nice to make the naming somewhat more idiomatic for Java.  On the other hand, it adds maintenance overhead if we expand the API in the future.

src/swig/CMakeLists.txt
```diff
@@ -71,3 +73,42 @@ if (GUILE_FOUND)
   set (GUILE_EXTENSION_PATH ${_REL_GUILE_EXTENSION_PATH})
   install (TARGETS nlopt_guile LIBRARY DESTINATION ${GUILE_EXTENSION_PATH})
 endif ()
+
+
+if (JNI_FOUND AND Java_FOUND AND SWIG_FOUND)
```
jschueller, line 78: build fails if guile is enabled because of this line, maybe it should be moved to the guile target only:
set_source_files_properties (nlopt.i PROPERTIES SWIG_FLAGS "-scmstub"):
```
swig error : Unrecognized option -scmstub
```

src/swig/nlopt.i
```diff
@@ -18,13 +19,18 @@ can be found at the NLopt web page: http://ab-initio.mit.edu/nlopt"
...
+#else
   %template(nlopt_doublevector) vector<double>;
+#endif
 };
 
 %ignore nlopt::opt::myfunc_data;
 %ignore nlopt::opt::operator=;
 
 // dont use the in-place version of get_initial_step
+%ignore nlopt_get_initial_step;
```
jschueller, line 33: why is this necessary ? isnt it a binding of the c++ interface, not the C one ?

CMakeLists.txt
```diff
@@ -42,6 +42,7 @@ option (NLOPT_PYTHON "build python bindings" ON)
 option (NLOPT_OCTAVE "build octave bindings" ON)
 option (NLOPT_MATLAB "build matlab bindings" OFF)
 option (NLOPT_GUILE "build guile bindings" ON)
+option (NLOPT_JAVA "build java bindings" ON)
```
jschueller, line 45: a new documentation section will be needed

src/swig/CMakeLists.txt
```diff
@@ -2,6 +2,8 @@
 if (POLICY CMP0078)
   cmake_policy(SET CMP0078 NEW)
 endif ()
+# clean up old generated source files before running SWIG, useful for Java
+set(UseSWIG_MODULE_VERSION 2)
```
jschueller, line 6: this is breaking the Python install rules because the locations are changed

CMakeLists.txt
```diff
@@ -344,6 +345,15 @@ if (NLOPT_GUILE)
   find_package (Guile)
 endif ()
 
+if (NLOPT_JAVA)
+  # we do not really need any component, only the main JNI target, but if the
+  # list of components is left empty, FindJNI defaults to "JVM AWT", and we
+  # specifically do not want to check for the AWT library that is not available
+  # on headless installations
+  find_package (JNI COMPONENTS JVM)
+  find_package (Java)
```
jschueller, line 354: we could specify the minimum version here instead of test/CMakeLists.txt

src/swig/CMakeLists.txt
kkofler: The Guile binding probably needs to do the same as what Java does for the `-package nlopt` flag, i.e., setting the global `SWIG_FLAGS` property before `swig_add_library` and unsetting it afterwards instead of using `set_source_files_properties`. I can make that change and update this pull request with it. The problem with using `set_source_files_properties` is that the same source `nlopt.i` is used for all target languages.

src/swig/nlopt.i
kkofler: Because SWIG was generating a binding for that function (using ugly `SWIGTYPE_p_*` types) without this line. I am not sure why that particular function was getting picked up and none of the other C interface function, but I guess the definition somehow ends up transitively included by `nlopt.hpp`.

src/swig/CMakeLists.txt
kkofler: Hmmm, I noticed that the Python tests were failing, but I was under the (possibly false) impression that they had already started failing recently before my changes. I am going to double-check where the failure comes from and look into fixing it as soon as possible either way, then updating the pull request.

CMakeLists.txt
kkofler: I will look into it.

CMakeLists.txt
kkofler: But as the comment and the commit message both clearly explain, the bindings will build with basically any version, probably as old as 1.4, maybe even 1.0. Only the test program has 3 lines that require 1.8. So it is wrong to require 1.8 here.

If you insist that the test needs to support everything the binding supports, I can change those to use explicit anonymous classes, but the result is ugly and I worry that users will copy&paste it even if they do not actually need support for ancient Java versions. (I expect users to use the test class as an example.)

src/swig/CMakeLists.txt
kkofler: But we may be talking about different things. What exactly is the issue with the Python bindings that you are seeing? Failing tests? `make install` failing? `make install` installing to a wrong location?

kkofler: I guess it is this hardcoded path in `${CMAKE_CURRENT_BINARY_DIR}` which is the issue for you:
```cmake
install (FILES ${CMAKE_CURRENT_BINARY_DIR}/nlopt.py DESTINATION ${INSTALL_PYTHON_DIR})
```
I will fix that ASAP and update the pull request (with this and the other fixes).

CMakeLists.txt
jschueller: looks like java 1.8 is pretty old (2014), maybe its a reasonable requirement

src/swig/CMakeLists.txt
```diff
@@ -71,3 +73,42 @@ if (GUILE_FOUND)
...
+  # on any changes to the SWIG interface code, they are not direct inputs.)
+  add_custom_command(OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/java_sources.txt
+                     COMMAND ${CMAKE_COMMAND}
+                             -DBINARY_DIR=${CMAKE_CURRENT_BINARY_DIR}
+                             -P ${CMAKE_CURRENT_SOURCE_DIR}/glob_java.cmake
+                     DEPENDS nlopt.i nlopt-exceptions.i nlopt-java.i
+                             nlopt_java_swig_compilation glob_java.cmake)
+
+  add_jar (nlopt_jar SOURCES @${CMAKE_CURRENT_BINARY_DIR}/java_sources.txt
+                     OUTPUT_NAME nlopt)
```
jschueller, line 113: we also need an install rule, into share/java for example

kkofler: Oops, true, I will fix that too.

src/swig/nlopt.i
kkofler: I have checked where that `nlopt_get_initial_step` function comes from: That function is defined in `nlopt.hpp`:
```c++
// convenience overloading for below (not in nlopt:: since has nlopt_ prefix)
inline nlopt_result nlopt_get_initial_step(const nlopt_opt opt, double *dx) {
      return nlopt_get_initial_step(opt, (const double *) NULL, dx);
}
```
so it is part of the C++ interface. It cannot be implemented in the C interface because it overloads the C function with differing arguments.

I am not sure why this overload is defined at all. The comment implies that is a helper used in the C++ inline methods below, but it is not actually used there. The `opt::get_initial_step` method calls the regular 3-argument `nlopt_get_initial_step` from the C interface instead.

In any case, it is in-place and therefore should be excluded from all bindings.

src/swig/CMakeLists.txt
kkofler: The `.java` sources need not be installed anywhere, the compiled `.class` files in the `.jar` are all that the Java compiler needs to build code depending on the library. The runtime then needs both the `.jar` and the JNI `.so`.

So only the install rule for the `.jar` is missing, I am adding that one now.

src/swig/CMakeLists.txt
kkofler: This should be fixed in the new revision (commit 32a4a64e7b087fb0cec719ccd794ed5a8fe2f585).

src/swig/CMakeLists.txt
kkofler: This should be fixed in the new revision (commit 32a4a64e7b087fb0cec719ccd794ed5a8fe2f585). Both the Python tests and `make install` work now.

src/swig/CMakeLists.txt
kkofler: This should be fixed in the new revision (commit 32a4a64e7b087fb0cec719ccd794ed5a8fe2f585). (I added the missing `install_jar` command.)

kkofler:
In the new revision (commit 32a4a64e7b087fb0cec719ccd794ed5a8fe2f585), I have fixed all the actual bugs you have found.

What is still unresolved is:
* the missing documentation – I will look into that some time next week.
* deciding how to handle the minimum Java version – I am not a huge fan of requiring a newer version than actually required, though I can see why you do not like allowing an older version than what the test requires, since it cannot be automatically tested. (Technically, one could build the binding with, e.g., JDK 1.6, then build and run the test with JDK 1.8, but not in one CMake run.) JDK 1.8 is the oldest we care about at my employer, so it should not be a big issue if it were required.

src/swig/CMakeLists.txt
```diff
@@ -71,3 +77,43 @@ if (GUILE_FOUND)
   set (GUILE_EXTENSION_PATH ${_REL_GUILE_EXTENSION_PATH})
   install (TARGETS nlopt_guile LIBRARY DESTINATION ${GUILE_EXTENSION_PATH})
 endif ()
+
```
jschueller, line 80: the guile install source must be changed too:
```
install (FILES ${CMAKE_CURRENT_BINARY_DIR}/nlopt.scm DESTINATION ${GUILE_SITE_PATH})
```

jschueller:
>     * deciding how to handle the minimum Java version
this was not a strong requirement, just a remark to make the code simpler, if you feel java < 1.6 will be useful you can leave the code as is

kkofler:
> It would be nice to make the naming somewhat more idiomatic for Java. On the other hand, it adds maintenance overhead if we expand the API in the future.

It turns out that SWIG supports global renaming directives that can do most of the work, as in:
```swig
// use Java naming conventions
%rename("%(camelcase)s", match="enum") "";
%rename("%(camelcase)s", match="class") "";
%rename("%(lowercamelcase)s", %$isfunction) "";
```
so I have implemented just that in the latest revision, commit bb97515d8024656d2fdeffedb4504defe0565c63.

I needed only a handful other changes (with 2 added `#ifdef`s in `nlopt.i`, the rest is centralized in `nlopt_java.i`) on top of those.

src/swig/CMakeLists.txt
kkofler: Should be fixed in the latest revision (commit 4ad857746f2e1b69c2fd610f3e6ecafa6356397c).

kkofler:
The latest revision (commit 7f09d014963930623a33cc54748d0536866c6a84) should hopefully also fix the CI failure on Windows. (The one on macOS was due to the Guile issue that I fixed in the previous revision.) The problem was that the classpath was using the wrong separator between paths on Windows. (It was hardcoded to `:`, most software on Windows uses `;` instead because `:` is used for the drive letter.)

kkofler:
Hmmmph, Windows is still unhappy, probably because CMake is munging the `;` somewhere.

Also, I noticed strict-aliasing warnings in the GNU/Linux CI run, `nloptJAVA_wrap.cxx` needs to be built with `-fno-strict-aliasing`, I need to look into that, too. (And of course the documentation that is still missing.)

kkofler:
> Hmmmph, Windows is still unhappy, probably because CMake is munging the `;` somewhere.

This and …

> Also, I noticed strict-aliasing warnings in the GNU/Linux CI run, `nloptJAVA_wrap.cxx` needs to be built with `-fno-strict-aliasing` (I have not seen those warnings locally)

… this should both be fixed in the latest revision (commit 4cab95daf76222bc89334e58b241ade6769f9a46).

kkofler:
The latest revision (commit 8c481a9b97c0af00a3c55f8cd6af1bc650e46cfc) now simplifies the syntax of the `DoubleVector` constructor so that you can write, e.g.:
```java
DoubleVector x0 = new DoubleVector(1.234, 5.678);
```
instead of:
```java
DoubleVector x0 = new DoubleVector(new double[]{1.234, 5.678});
```
This is done by using the varargs syntax `double... initialElements` introduced in Java 1.5 instead of the array syntax `double[] initialElements`. Note that, with Java varargs, you can still pass an explicitly constructed array (so both snippets above are now valid), useful if you already have an array.

This raises the minimum Java version for the Java binding to 1.5. (The test program still requires 1.8.) For some perspective, Java 1.5 was released in 2004 and reached EOL in 2009, Java 1.4 (which is no longer supported) was released in 2002 and reached EOL in 2008.

SWIG is unfortunately still unconditionally using the legacy Java 1.4 syntax in 2024 and requiring a manual replacement in the output file to fix this. (The `[]` is hardcoded in a `%proxycode` block in `std_vector.i`, so it cannot even be customized via `%typemap`.) Backwards compatibility of the generated output is not even an argument, because as I mentioned above, varargs also accept an explicit array, so it really seems to be about not wanting to require Java 1.5 in SWIG. So I added the replacement to the `glob_java.cmake` snippet that runs at the right spot of the build and is needed anyway.

kkofler:
The Java tests are still failing on Windows, and I do not know why it (still) does not find the class in the classpath there. I do not have a Windows computer to test with locally.

kkofler:
There are actually several issues with the Windows CI build, e.g., it explicitly installs NumPy with `pip install numpy`, but then CMake does not find it. There is also some peculiar behavior by MSBuild, or the CMake support for it: There are a bunch of errors occurring during the build and it just proceeds, triggering more errors. And it seems to also not honor ordering dependencies.

jschueller:
I fixed the numpy detection on windows, but dlls are still manually copied to the swig binary dir, so if the target dir changes you will have to modify the destination

test/CMakeLists.txt
```diff
@@ -63,9 +63,13 @@ foreach (algo_index RANGE 28) # 42
...
 
 if (Python_FOUND AND NUMPY_FOUND AND (SWIG_FOUND OR (EXISTS ${PROJECT_SOURCE_DIR}/src/swig/nlopt-python.cpp)))
-  set (PYINSTALLCHECK_ENVIRONMENT "LD_LIBRARY_PATH=${PROJECT_BINARY_DIR}/src/swig"
-                                  "PYTHONPATH=${PROJECT_BINARY_DIR}/src/swig"
-    )
+  if (CMAKE_HOST_SYSTEM_NAME MATCHES Windows)
+    set (PYINSTALLCHECK_ENVIRONMENT "PYTHONPATH=${PROJECT_BINARY_DIR}/src/swig/python;${PROJECT_BINARY_DIR}/src/swig")
+  else()
+    set (PYINSTALLCHECK_ENVIRONMENT "LD_LIBRARY_PATH=${PROJECT_BINARY_DIR}/src/swig"
+                                    "PYTHONPATH=${PROJECT_BINARY_DIR}/src/swig/python:${PROJECT_BINARY_DIR}/src/swig"
```
jschueller, line 70: it does not hurt to set LD_LIBRARY_PATH on windows, that avoids the unnecessary if/else

src/swig/glob_java.cmake
```diff
@@ -0,0 +1,13 @@
...
+file(GLOB JAVA_SOURCES ${BINARY_DIR}/java/nlopt/*.java)
+list(JOIN JAVA_SOURCES "\n" JAVA_SOURCES_LINES)
+file(WRITE ${BINARY_DIR}/java_sources.txt ${JAVA_SOURCES_LINES})
+
+# SWIG hardcodes non-vararg initial elements for std::vector wrappers,
+# probably to support Java versions older than 1.5. We do not really care
+# about supporting a Java that old, so fix the generated code.
+file(READ ${BINARY_DIR}/java/nlopt/DoubleVector.java FILE_CONTENTS)
+string(REPLACE "double[] initialElements" "double... initialElements" FILE_CONTENTS "${FILE_CONTENTS}")
+file(WRITE ${BINARY_DIR}/java/nlopt/DoubleVector.java "${FILE_CONTENTS}")
```
jschueller, line 13: do you mind opening a bug report at SWIG for that ? maybe it could be resolved on the SWIG side with an option or something

CMakeLists.txt
kkofler: The latest revision (commit fe30d98be818f511b5e6759dde8bc56297c1da3a) adds the missing documentation.

kkofler:
The latest revision (commit fe30d98be818f511b5e6759dde8bc56297c1da3a) adds the missing documentation that was requested in the review.

It also fixes the exception handling to work as in Python (noticed while adapting the documentation): Runtime exceptions thrown by the Java callbacks now get rethrown as is (with the original class, message, and stack trace) rather than as a `ForcedStopException` that lost all the original information. (Another way in Java would be to set the original exception as a `cause` in the `ForcedStopException`, but that is mostly useful when wrapping checked exceptions into a runtime exception or into another checked exception class. It is unnecessary for runtime exceptions, where we can avoid wrapping the exception to begin with.)

jschueller:
seems there is a conflict with master (I fixed the numpy detection on the CI and improved it on the cmake side)

test/CMakeLists.txt
kkofler: Actually, since you bumped the minimum CMake version in master from 3.13 to 3.15, we can now simply use https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:SHELL_PATH with semicolon-separated paths to let CMake automatically pick the proper separator, a feature that was introduced in CMake 3.14.

kkofler: Done in the latest revision (commit 427ca219a7ebd670513b8f0a4229facddb7dd229).

kkofler:
What I still do not fully understand is the reason for the Windows CI failure. As far as I can tell, the first error is this:
```
  Generating java_sources.txt
  CMake Error at glob_java.cmake:11 (file):
    file failed to open for reading (No such file or directory):
  
      D:/a/nlopt/nlopt/src/swig/java/nlopt/DoubleVector.java
```
which means that the `add_custom_command` rule for `java_sources.txt` executes too early, before the SWIG run that generates those `.java` files (also the other ones that should end up listed in `java_sources.txt`). I believe that the remaining errors are probably followup errors.

The rule has a `DEPENDS` on `nlopt_java_swig_compilation`. On the "UNIX Makefiles" target, this appears to be sufficient to get the correct ordering, but not on the MSBuild target.

kkofler:
The latest revision (commit f4a1e4efcc3aa3bb056ba53514911795ccfe2889) attempts to fix this, but I cannot test it on Windows, so I can only hope for the best.

What this does is add a dependency on `"${swig_generated_file_fullname}"`, which is a variable set by `UseSWIG.cmake` to the main SWIG output file (e.g., in this case, `src/swig/nloptJAVA_wrap.cxx`), since, looking at that file, it looks like `nlopt_java_swig_compilation` is only used for Makefile targets. In addition, it also makes this the `MAIN_DEPENDENCY`, because the documentation for `add_custom_command` says that doing that "also suggests to Visual Studio generators where to hang the custom command." So I really hope the ordering will be right now.

kkofler:
Hmmmph, the added `MAIN_DEPENDENCY` is definitely not good, it now generates an empty `libnloptjni.so` for me! Looks like I forgot to run the tests.

kkofler:
So I have now made that dependency a regular `DEPENDS`. No idea whether it will be good enough for Windows, but at least it does not break GNU/Linux.

Using it as a `MAIN_DEPENDENCY` violates this constraint: "Each source file may have at most one command specifying it as its main dependency." that I had overlooked.

kkofler:
The latest revision (commit 819ee46e6c125f0dd20b9c408dca91dcd207ed29) fixes the build with CMake < 3.24, e.g, I tried it in a CentOS 7 chroot with the `cmake3` from EPEL, which is 3.17.5. (The regular `cmake` on the ancient CentOS 7 is 2.8.12.2 and too old to build NLopt altogether.)

doc/docs/NLopt_Java_Reference.md
```diff
@@ -0,0 +1,423 @@
...
+
+```java
+int major = NLopt.versionMajor();
+int minor = NLopt.versionMinor();
+int bugfix = NLopt.versionBugfix();
+```
+
+For example, NLopt version 3.1.4 would return `major=3`, `minor=1`, and `bugfix=4`.
+
+
```
jschueller, line 423: the documentation looks great!

test/CMakeLists.txt
```diff
@@ -64,8 +64,8 @@ endforeach ()
 
 if (Python_NumPy_FOUND AND (SWIG_FOUND OR (EXISTS ${PROJECT_SOURCE_DIR}/src/swig/nlopt-python.cpp)))
   set (PYINSTALLCHECK_ENVIRONMENT "LD_LIBRARY_PATH=${PROJECT_BINARY_DIR}/src/swig"
-                                  "PYTHONPATH=${PROJECT_BINARY_DIR}/src/swig"
-    )
+                                  "PYTHONPATH=$<SHELL_PATH:${PROJECT_BINARY_DIR}/src/swig/python;${PROJECT_BINARY_DIR}/src/swig>"
+      )
```
jschueller, line 69: now the location for the manual dll copy must be changed too (probably not great):
```
diff --git a/.github/workflows/build.yml b/.github/workflows/build.yml
index 4abae4d..2ba7900 100644
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -41,7 +41,7 @@ jobs:
           pip install numpy
           cmake -DCMAKE_INSTALL_PREFIX=%cd:\=/%/install -DNLOPT_TESTS=ON .
           cmake --build . --config Release --target install
-          xcopy %cd%\install\bin\nlopt.dll %cd%\src\swig
-          xcopy %cd%\install\Lib\site-packages\_nlopt*.pyd %cd%\src\swig
+          xcopy %cd%\install\bin\nlopt.dll %cd%\src\swig\python
+          xcopy %cd%\install\Lib\site-packages\_nlopt*.pyd %cd%\src\swig\python
           dir /p %cd%\src\swig
           ctest -C Release --output-on-failure --timeout 100
```
I guess the same thing is missing that makes the java tests fails too on windows

src/swig/CMakeLists.txt
```diff
@@ -58,9 +66,54 @@ if (GUILE_FOUND)
...
+                     COMMAND ${CMAKE_COMMAND}
+                             -DBINARY_DIR=${CMAKE_CURRENT_BINARY_DIR}
+                             -P ${CMAKE_CURRENT_SOURCE_DIR}/glob_java.cmake
+                     DEPENDS "${swig_generated_file_fullname}"
+                             nlopt.i nlopt-exceptions.i nlopt-java.i
+                             nlopt_java_swig_compilation glob_java.cmake)
+
+  add_jar (nlopt_jar SOURCES @${CMAKE_CURRENT_BINARY_DIR}/java_sources.txt
+                     OUTPUT_NAME nlopt)
+  install_jar (nlopt_jar ${CMAKE_INSTALL_DATADIR}/java)
```
jschueller, line 118: the .java files must be installed too, maybe with the "install(CODE" signature as the list must be read from java_sources.txt via another cmake script

doc/docs/NLopt_Java_Reference.md
kkofler: > the documentation looks great!

Thanks!

test/CMakeLists.txt
kkofler: Yes, the Java tests also fail on Windows for the same reason (a DLL that was not found). I will look into it. Unfortunately, in my experience, DLL search paths are a particular pain on Windows: It supports neither RPATH/RPATH nor LD_LIBRARY_PATH. Sometimes, it finds DLLs in PATH, except when it does not. And sometimes it honors search paths set somewhere in the registry, except when it does not. Copying the DLL multiple times is often the only reliable way. I will tweak the xcopy lines and hope I get them right.

kkofler: This should be fixed in the latest revision (commit 062dd630c17cd0ad48a90a889f8cf737c7fac2b7): The Python stuff gets copied to the new location. For Java, `java.library.path` seems to work to get Java to find `nloptjni.dll`, but then `nlopt.dll` (on which `nloptjni.dll` depends) was not found, so I now copy `nlopt.dll` to where `nloptjni.dll` is found.

I also fixed the installed location of `nloptjni.dll` (should be in `bin`, not `lib`, same as `nlopt.dll`), but the tests are not using it from there, but from the build tree.

src/swig/glob_java.cmake
kkofler: Filed swig/swig#3085.

src/swig/CMakeLists.txt
```diff
@@ -58,9 +66,56 @@ if (GUILE_FOUND)
...
+  # swig_add_module is deprecated
+  # OUTPUT_DIR is ${CMAKE_CURRENT_BINARY_DIR}/java/ + the -package above (with
+  # any '.' replaced by '/'). It must also match the GLOB in glob_java.cmake.
+  swig_add_library (nlopt_java LANGUAGE java SOURCES nlopt.i
+                               OUTPUT_DIR ${CMAKE_CURRENT_BINARY_DIR}/java/nlopt
+                               OUTFILE_DIR ${CMAKE_CURRENT_BINARY_DIR})
+
+  set (CMAKE_SWIG_FLAGS)
+
+  swig_link_libraries (nlopt_java ${nlopt_lib})
```
jschueller, line 93: swig_link_libraries is deprecated, use target_link_libraries instead

src/swig/CMakeLists.txt
```diff
@@ -58,9 +66,56 @@ if (GUILE_FOUND)
...
+
+  set_target_properties (nlopt_java PROPERTIES OUTPUT_NAME nloptjni)
+  if (CMAKE_CXX_COMPILER_ID MATCHES "Clang|GNU")
+    set_target_properties (nlopt_java PROPERTIES
+                           COMPILE_OPTIONS "-fno-strict-aliasing")
+  endif ()
+
+  install (TARGETS nlopt_java RUNTIME DESTINATION ${NLOPT_INSTALL_BINDIR}
+                              LIBRARY DESTINATION ${NLOPT_INSTALL_LIBDIR}
+                              ARCHIVE DESTINATION ${NLOPT_INSTALL_LIBDIR})
```
jschueller, line 104: that's curious, nloptjni looks still installed in /lib:
```
 Installing: D:/a/nlopt/nlopt/install/lib/nloptjni.dll
```

.github/workflows/build.yml
```diff
@@ -41,7 +41,8 @@ jobs:
           pip install numpy
           cmake -DCMAKE_INSTALL_PREFIX=%cd:\=/%/install -DNLOPT_TESTS=ON .
           cmake --build . --config Release --target install
-          xcopy %cd%\install\bin\nlopt.dll %cd%\src\swig
-          xcopy %cd%\install\Lib\site-packages\_nlopt*.pyd %cd%\src\swig
+          xcopy %cd%\install\bin\nlopt.dll %cd%\src\swig\python
+          xcopy %cd%\install\Lib\site-packages\_nlopt*.pyd %cd%\src\swig\python
+          xcopy %cd%\install\bin\nlopt.dll %cd%\src\swig\Release
```
jschueller, line 46: looks like the right path
D:\a\nlopt\nlopt\src\swig\Release\nloptjni.dll
maybe another java dll is missing ?

src/swig/CMakeLists.txt
kkofler: That one was copy&pasted (cargo-culted) from the existing Guile binding. I am fixing it in both.

src/swig/CMakeLists.txt
kkofler: That is because it is of MODULE type. See https://cmake.org/cmake/help/book/mastering-cmake/chapter/Install.html#installing-targets . So if we want to install it to `bin`, we have to set the `LIBRARY DESTINATION` conditionally for Windows.

src/swig/CMakeLists.txt
kkofler: Both now use `target_link_libraries` in the latest revision (commit 8cd6dad0b8c9fbec9a7282a5531674bf744bd9a3).

src/swig/CMakeLists.txt
kkofler: The latest revision (commit 8cd6dad0b8c9fbec9a7282a5531674bf744bd9a3) now sets the destination for `nlopt_java` correctly.

.github/workflows/build.yml
kkofler: Unfortunately, this part (which is probably the last error making this fail in the CI) is still unclear to me. `nloptjni.dll` should not have any dependencies other than `nlopt.dll`. JNI libraries normally do not link to any Java library. In any case, when I build NLopt with cross-MinGW, it does not have any extraneous dependency. (Depending on the build flags, there can be dependencies on MinGW DLLs, but this should not be the issue for the MSVC build either.)

It sounds more to me like `nlopt.dll` is not being found even if it is in the same directory as `nloptjni.dll`.

jschueller (approved):
great! 
just a rebase to master and it should pass CI

kkofler:
Rebased now.

kkofler:
Rebased to the current `master`, the trivial conflict in `README.md` fixed.

Now that this is approved, is this going to be merged any time soon?

jschueller:
hello @stevengj,
This can be merged, what do you think ?

PR #580, opened 2024-12-20
C++: Allow disabling exceptions in opt::optimize.

kkofler:
Call opt.set_exceptions_enabled(false) before opt.optimize to avoid throwing an exception. The caller is then responsible for checking the returned result or opt.last_result(). The advantage is that, for the SWIG-friendly overload, the returned best point (which can be quite good in, e.g., the roundoff_limited case) is not lost (but then the caller MUST check opt.last_result() to know whether anything has gone wrong). The default is exceptions_enabled=true, so by default, nothing changes.

src/api/nlopt-in.hpp (opt::exceptions_enabled): New private member variable.
(opt::opt (all overloads)): Initialize it (default it to true, copy it in the copy constructor).
(opt::operator=): Copy it.
(opt::optimize): Only call mythrow if exceptions_enabled. (get_exceptions_enabled): New public inline getter. (set_exceptions_enabled): New public inline setter.

### DIFF
--- a/src/api/nlopt-in.hpp
+++ b/src/api/nlopt-in.hpp
@@ -246,26 +246,27 @@ namespace nlopt {
       }
     }
 
+    bool exceptions_enabled;
     result last_result;
     double last_optf;
     nlopt_result forced_stop_reason;
 
   public:
     // Constructors etc.
-    opt() : o(NULL), xtmp(0), gradtmp(0), gradtmp0(0),
+    opt() : o(NULL), xtmp(0), gradtmp(0), gradtmp0(0), exceptions_enabled(true),
 	    last_result(nlopt::FAILURE), last_optf(HUGE_VAL),
 	    forced_stop_reason(NLOPT_FORCED_STOP) {}
     ~opt() { nlopt_destroy(o); }
     opt(algorithm a, unsigned n) :
       o(nlopt_create(nlopt_algorithm(a), n)),
-      xtmp(0), gradtmp(0), gradtmp0(0),
+      xtmp(0), gradtmp(0), gradtmp0(0), exceptions_enabled(true),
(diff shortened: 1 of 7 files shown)


src/api/nlopt-in.hpp
```diff
@@ -598,6 +603,10 @@ namespace nlopt {
       get_initial_step(x, v);
       return v;
     }
+
+    // exceptions in opt::optimize:
+    bool get_exceptions_enabled() const { return exceptions_enabled; }
+    void set_exceptions_enabled(bool enable) { exceptions_enabled = enable; }
```
jschueller, line 609: these must be documented

kkofler: Followup commit e94cd1f3d1d106cfc603713b72bed16a68da862c adds the missing documentation for the affected languages (C++, Guile, Python). (Note that the Java documentation will need basically the same changes as the Python one, but the Java documentation is not in `master` yet and I do not want the pull requests to depend on each other.)

kkofler: I have the Java documentation in my fork as commit d29bae391e6d92d632b33b01876e409d354bca12, I can cherry-pick it somewhere when one or both of the pull requests are merged.

jschueller, line 609: we should probably add a new c++ test for this

jschueller (approved):
ok for me

jschueller:
is there something more to do for the java rebase ?
or maybe just the java reference doc ?

kkofler:
Just the doc, give me a few minutes, I am going to take care of the rebase right now.

kkofler:
Done. This is now rebased and the Java documentation commit added. Should be ready for merge now.

jschueller:
thanks a lot, I merged everything

Issue #121, opened 2017-05-18
gradient based stopping criterion


mithodin:
For gradient based optimization algorithms, a stopping criterion based on the gradient would be useful. One could, for example, check that the norm of the gradient is equal to zero up to some user-specified tolerance.

stevengj:
Yes, this is certainly possible (with modifications to handle active constraints).  One annoyance is that the norm of the gradient is dimensionful, and that makes it especially difficult for users to come up with a reasonable tolerance.

helske:
Has there been any plans to implement this?

Benoit-F-Q:
I agree with helske, but the gradient tolerance `tolg` is not settable by the user.

In our team, we patched the hardcoded value.
I wish to share something more useable (open a PR), but I wonder where to add a "luksan_tolg" setter which would be meaningless for a majority of algorithms

PR #585, opened 2025-01-20
configurable tolerance for gradient stopping condition (#121)

Benoit-F-Q:
https://github.com/stevengj/nlopt/issues/121

Luksan algorithms (BFGS, troncated Newton, Shifted variable-metric) use a hard-coded value for their gradient stopping condition.

The tolerance is exposed to the API, with current value as default.

### DIFF
--- a/src/api/optimize.c
+++ b/src/api/optimize.c
@@ -556,6 +556,9 @@ static nlopt_result nlopt_optimize_(nlopt_opt opt, double *x, double *minf)
     stop.ftol_abs = opt->ftol_abs;
     stop.xtol_rel = opt->xtol_rel;
     stop.xtol_abs = opt->xtol_abs;
+    stop.tolg = (opt->tolg > 0.)
+        ? opt->tolg
+        : 1e-8; /* SGJ: was 1e-6, but this sometimes stops too soon */
     stop.x_weights = opt->x_weights;
     opt->numevals = 0;
     stop.nevals_p = &(opt->numevals);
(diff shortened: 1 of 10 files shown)


jschueller:
it seems we should use the nlopt_set_param api instead of adding a silent struct entry, to make it accessible for binding languages (eg Python)

src/api/optimize.c
```diff
@@ -735,7 +735,7 @@ static nlopt_result nlopt_optimize_(nlopt_opt opt, double *x, double *minf)
     case NLOPT_LD_TNEWTON_PRECOND:
     case NLOPT_LD_TNEWTON_PRECOND_RESTART:
 #ifdef NLOPT_LUKSAN
-        return luksan_pnet(ni, f, f_data, lb, ub, x, minf, &stop, opt->vector_storage, 1 + (algorithm - NLOPT_LD_TNEWTON) % 2, 1 + (algorithm - NLOPT_LD_TNEWTON) / 2);
+        return luksan_pnet(ni, f, f_data, lb, ub, x, minf, &stop, opt->vector_storage, 1 + (algorithm - NLOPT_LD_TNEWTON) % 2, 1 + (algorithm - NLOPT_LD_TNEWTON) / 2, opt->get_param("tolg", 0.));
```
jschueller, line 738: maybe we could set the default value here and simplify the rest instead of using 0 which only means use the default that will be set later

Benoit-F-Q: I considered your idea, what stopped me from choosing it was :
* using a default here won't cover the case of the user entering an invalid value (negative or null is currently switched back to default)
* I find that the default value belongs more legitimately to the algorithm than to the generic caller
* it is technically cumbursome to define a variable inside a `case` (you need braces)

None of these items is a real issue, their sum oriented me toward my current proposal, but I totally agree if you prefer the other way.

jschueller: ok

Benoit-F-Q: (shame on me!) my apologies...

jschueller (approved):

PR #586, opened 2025-01-23
CMake: Call project command before anything else

StephanOostveen:
Call the cmake project command before setting up any variables or calling any other cmake command to ensure that the language is properly set etc. Otherwise this can lead to weird behaviour such as seen here: https://gitlab.kitware.com/cmake/cmake/-/issues/22900


jschueller:
-1
this line is here to initialize the build type, I dont think that it poses problem by itself and that we can do that in another way

Issue #584, opened 2025-01-05
Commit f4e74eb broke NLopt.jl (and any wrappers which do not pull in a fresh .h)


odow:
### What happened?

This commit https://github.com/stevengj/nlopt/commit/f4e74eb164d4ef650a339ef3eb4b34df177ee6e5 dropped the `NLOPT_LD_LBFGS_NOCEDAL` enum.

This broke NLopt.jl because we hard-code the enums in Julia code:
https://github.com/jump-dev/NLopt.jl/blob/bdaa35d626b2168a77251224a5d1d7a2a9677358/src/libnlopt.jl#L19-L31

We can fix in our code with a conditional that check the version of NLopt, but I wonder if you would consider re-adding the (admittedly, unused) enum as a bug fix.

### How to reproduce the issue?

```shell
N/A.
```

### Version

2.9.0

### Operating System

all

### Installation media

other

jschueller:
I'd prefer not to, it can be misleading for users to be able to select an algorithm that's not actually available

odow:
Here's the fix I needed: https://github.com/jump-dev/NLopt.jl/pull/251/files

I assume other third-party wrappers will need to do something similar?

jschueller:
I guess it depends if they have to explicit all the enum values

stevengj:
(The reason that `NLOPT_LD_LBFGS_NOCEDAL` exists is that a very early version of NLopt included support for Nocedal's LBFGS implementation.  I removed it because it was under the restrictive ACM TOMS copyright terms.   Subsequently, at Nocedal's request, ACM relaxed the terms for that code to be open source BSD-like, IIRC, so potentially we could add it back.)

aadler:
This broke [nloptr](https://github.com/astamm/nloptr) as well, for what it is worth ([see here](https://github.com/astamm/nloptr/issues/173)). I reached out to the administrator for [Rtools for Windows](https://stat.ethz.ch/pipermail/r-sig-windows/2025q1/000120.html), and he will freeze nlopt at 2.8.0 until we can fix this as well. May I request that in the future, if you are considering a change that will break the API, that you let downstream users know prior? Thanks!

odow:
Just re-iterating my support for re-introducing the enum value as a bug-fix. (Even though it'll break the work-around in NLopt, but I can revert that.)

PR #588, opened 2025-02-04
release 2.10

stevengj:
Bump versions for 2.10.

@jschueller, any reason why we shouldn't release this now (because of #584)?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,15 +21,15 @@ project (nlopt)
 #==============================================================================
 # version
 set (NLOPT_MAJOR_VERSION "2")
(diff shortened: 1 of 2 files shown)


jschueller:
its all ready I think

jschueller:
it doesnt look like the documentation has been updated, any idea why ?

stevengj:
😕 not sure

stevengj:
readthedocs.io has an error message:

> Config file not found at default path: The required .readthedocs.yaml configuration file was not found at repository's root. Learn how to use this file in our [configuration file tutorial](https://docs.readthedocs.io/en/stable/config-file/index.html).

PR #591, opened 2025-02-16
print evaluation count in t_tutorial.cxx

stevengj:
For better diagnostics.

e.g. with the default algorithm, the output is now:
```
found minimum at f(0.333333,0.296296) = 0.5443310497 after 48 evaluations
```


jschueller:
seems AUGLAG runs  forever here

stevengj:
Probably the `local_opt` convergence criteria need to be fixed for AUGLAG.

PR #594, opened 2025-02-21
Add an internal parameter "outer_mineval_xftol" for mma/ccsaq.

madosqr:
In mma/ccsaq, when the outer iteration step k is less than "outer_mineval_xftol", the stopping criteria of xtol and ftol are disabled. This is quite useful in some optimization problems where there is a "startup" process during which x and f change slowly. The default value of "outer_mineval_xftol" is 0, namely the program behavior is the same with that of the current version.

	modified:   src/algs/mma/ccsa_quadratic.c
	modified:   src/algs/mma/mma.c
	modified:   src/algs/mma/mma.h
	modified:   src/api/optimize.c

### DIFF
--- a/src/algs/mma/ccsa_quadratic.c
+++ b/src/algs/mma/ccsa_quadratic.c
@@ -218,7 +218,7 @@ nlopt_result ccsa_quadratic_minimize(
      double *x, /* in: initial guess, out: minimizer */
      double *minf,
      nlopt_stopping *stop,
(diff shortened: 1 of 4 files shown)


madosqr:
Now "outer_mineval_xftol" is documented in NLopt_Algorithms.md
The indentation is corrected.

jschueller:
ok for me

PR #595, opened 2025-02-24
Use find_library to check if libm is available

michalsc:
Use find_library CMake function to detect presence of m-library - this version prevents different use of M_LIBRARY variable which was eventually cached.

This change solves conflict issues when NLopt is built as a subproject together with other subprojects in the larger product. Many libraries are using M_LIBRARY variable to search for presence of libm, but they use find_library for that. According to CMake docs the find_library() function sets cached variable.

Another solution would be to reset M_LIBRARY to empty value if libm was not found:

```
check_library_exists ("m" sqrt "" HAVE_LIBM)
if (HAVE_LIBM)
  set (M_LIBRARY m)
  set (LIBS_PRIVATE "-l${M_LIBRARY}")
else()
  set (M_LIBRARY "")
endif()
```

please apply either my patch or the suggested correction above, whichever you are comfortable with :)


jschueller:
done

Issue #604, opened 2025-03-25
[R wrapper] Compiled code should not call entry points which might terminate R


astamm:
I am the maintainer of the **nloptr** R package which ports the *nlopt* library for the R community.

With the upcoming R 4.5 release, warnings appeared of the following form:

```bash
File ‘nloptr/libs/nloptr.so’:
    Found ‘_ZSt4cerr’, possibly from ‘std::cerr’ (C++)
      Object: ‘nlopt/lib/libnlopt.a’
    Found ‘_ZSt4cout’, possibly from ‘std::cout’ (C++)
      Object: ‘nlopt/lib/libnlopt.a’
    Found ‘__printf_chk’, possibly from ‘printf’ (C)
      Object: ‘nlopt/lib/libnlopt.a’
    Found ‘abort’, possibly from ‘abort’ (C)
      Object: ‘nlopt/lib/libnlopt.a’
    Found ‘exit’, possibly from ‘exit’ (C)
      Object: ‘nlopt/lib/libnlopt.a’
    Found ‘putchar’, possibly from ‘putchar’ (C)
      Object: ‘nlopt/lib/libnlopt.a’
    Found ‘puts’, possibly from ‘printf’ (C), ‘puts’ (C)
      Object: ‘nlopt/lib/libnlopt.a’
    Found ‘stderr’, possibly from ‘stderr’ (C)
      Object: ‘nlopt/lib/libnlopt.a’

Compiled code should not call entry points which might terminate R nor
  write to stdout/stderr instead of to the console, nor use Fortran I/O
  nor system RNGs nor [v]sprintf.
```

For instance I see a lot of usage of `stderr` and `fprintf` in `cobyla.c` but not in other algorithms.

Do you think these can be adapted / commented out to avoid their use or should I just patch nlopt sources to be suitable for CRAN rules? Thanks!

PR #608, opened 2025-04-17
Ease bindings to other languages providing internal interface for relaying messages and errors and taking care of sanitizers issues

astamm:
This is useful for providing *nlopt* to the R community while satisfying CRAN requirements.

Specifically, `CMakeLists.txt` gains  the option `USE_R_COMPATIBILITY`:

```cmake
option (USE_R_COMPATIBILITY "compile with R compatibility functions only" OFF)
```

which, when turned ON, triggers the following definition:

```cmake
if (USE_R_COMPATIBILITY)
  add_definitions(-DCRAN_COMPATIBILITY)
endif ()
```

The variable `CRAN_COMPATIBILITY` is then used in source files to prevent the use of functions forbidden by CRAN such as `stderr`, `abort`, `printf`.

This PR provides the necessary fixes for the C API. There are also known uses of `std::cerr` and `std::cout` notably in the STOGO implementation that could also be avoided using `CRAN_COMPATIBILITY`. It might later be good to make also the C++ API *R-safe*.


astamm:
@aadler I noticed you forked *nlopt* as well last month. Just pinging you into this PR in case you had the same idea.

astamm:
This PR should address #604.

astamm:
Last commit brings a new `FindR.cmake` file to properly set notably `R_INCLUDE_DIR` so that, instead of disabling verbosity and error handling, we can replace the R-incompatible functions with their R-compatible counterparts (e.g. `Rprintf()`, `REprintf()`, `Rf_error()`).

astamm:
So instead of a simple flag option of the type `NLOPT_R = ON/OFF`, the `CMakeLists,.txt` now gains a variable `NLOPT_R_BINDIR` which is empty by default:

```cmake
set (NLOPT_R_BINDIR "" CACHE STRING "Directory containing the R executable")
```

When set to the directory containing the R executable, it triggers:

```cmake
if (NOT "${NLOPT_R_BINDIR}" STREQUAL "")
  find_package(R REQUIRED)
  target_include_directories(${nlopt_lib} PRIVATE ${R_INCLUDE_DIR})
  target_compile_definitions(${nlopt_lib} PRIVATE NLOPT_R)
endif ()
```

which makes all headers from R's C API available to *nlopt* code and defines `NLOPT_R` which can be used to switch from e.g. `printf()` to `Rprintf()` in the source code.

aitap (commented):
I understand that the perfect is the enemy of the good, but since it seems that `nlopt` needs to go over its `[f]printf`s anyway, wouldn't it be better served by a layer of indirection? The optimisation methods would call an internal interface `nlopt_log(opt, "format", args...)`, and the caller of the library would supply appropriate callbacks (such as `Rprintf`) to make sure that the user sees the output. The defaults for the callback (`printf`) could be disabled at compile time. This would avoid duplication. (Imagine if another interface needs to introduce an `#ifndef NLOPT_SOME_OTHER_LANGUAGE` in the same place for the same reason.)

Calling `R[E]printf` is not exactly safe, by the way. For example, on `Rgui.exe`, this checks for interrupts and would `longjmp()` out of the calling frame if an interrupt is found. This will at least leak memory, or [wreak complete havoc](https://github.com/tnagler/RcppThread/issues/79) if code ever becomes multi-threaded.

astamm:
> I understand that the perfect is the enemy of the good, but since it seems that `nlopt` needs to go over its `[f]printf`s anyway, wouldn't it be better served by a layer of indirection? The optimisation methods would call an internal interface `nlopt_log(opt, "format", args...)`, and the caller of the library would supply appropriate callbacks (such as `Rprintf`) to make sure that the user sees the output. The defaults for the callback (`printf`) could be disabled at compile time. This would avoid duplication. (Imagine if another interface needs to introduce an `#ifndef NLOPT_SOME_OTHER_LANGUAGE` in the same place for the same reason.)

I agree with you. However, there are already existing bindings for many languages. I do not know how these have been achieved in the past. A reflection on this together with a layer of indirection to accommodate any past and future languages (maybe Julia aficionados) could be a long-term project on which we could work as a small group.

On a shorter term, **nloptr** does not pass CRAN checks anymore because in the case of it being built from included sources (which are the latest 2.10), CRAN now checks for calls to these functions (`cerr`, `cout` and co) and issues warning if these are used. Could we consider acceptable the currently proposed solution since it does not alter anyone's code unless (s)he specifically turn `USE_R_COMPATIBILITY` on?

Of course, I will answer the questions you raised in your review first.

> Calling `R[E]printf` is not exactly safe, by the way. For example, on `Rgui.exe`, this checks for interrupts and would `longjmp()` out of the calling frame if an interrupt is found. This will at least leak memory, or [wreak complete havoc](https://github.com/tnagler/RcppThread/issues/79) if code ever becomes multi-threaded.

You are right indeed. I think posit folks have come up in their [**cpp11**](https://cpp11.r-lib.org) package with ways to deal with this. I will investigate.

cmake/FindR.cmake
```diff
@@ -0,0 +1,99 @@
...
+find_file(
+          R_LIBRARY
+          NAMES R.dll
+          PATHS ${R_BIN_DIR}
+          NO_DEFAULT_PATH
+)
+else()
+set(R_PATH_LOC "${R_HOME_TEXT}/lib")
+MESSAGE(STATUS "R_PATH_LOC = ${R_PATH_LOC}")
+find_library(
```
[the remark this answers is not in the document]
astamm: Thanks for this feedback. Do you know how I could improve the `FindR.cmake` file to circumvent this issue?

src/algs/direct/direct-internal.h
```diff
@@ -17,7 +17,11 @@ typedef int integer;
 typedef double doublereal;
 typedef direct_objective_func fp;
 
+#ifndef NLOPT_R
 #define ASRT(c) if (!(c)) { fprintf(stderr, "DIRECT assertion failure at " __FILE__ ":%d -- " #c "\n", __LINE__); exit(EXIT_FAILURE); }
+#else
+#define ASRT(c) if (!(c)) { Rf_error("DIRECT assertion failure at " __FILE__ ":%d -- " #c "\n", __LINE__); }
```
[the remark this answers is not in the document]
astamm: Originally I wanted to avoid the need to call for **Rcpp** or **cpp11** functions within *nlopt* codebase because this would make a forced choice for R users to use one of them.

But I think it could make sense to force the use of **cpp11** as I recall they carefully dealt with `longjmp`s.

src/api/nlopt-internal.h
```diff
@@ -107,7 +107,9 @@ extern "C" {
 
     extern const char *nlopt_set_errmsg(nlopt_opt opt, const char *format, ...)
 #ifdef __GNUC__
+#ifndef NLOPT_R
         __attribute__ ((format(printf, 2, 3)))
```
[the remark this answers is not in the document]
astamm: Probably not needed indeed.

astamm:
@aitap is this something more along the lines of what you had in mind? And to provide an example with the R language, I updated the sandbox package {nloptrbundled}: https://github.com/astamm/nloptrbundled.

Now there is no mention of R and the message and error relays are handled by downstream languages which seek to provide a binding.

The R package example passes all CRAN checks on windows, Mac and linux successfully.
The nlopt code base is not polluted with `#ifdef` everywhere, and it should ease bindings for other languages.

Let me know if this can be considered for the main branch.

src/algs/auglag/auglag.c
```diff
@@ -1,290 +1,338 @@
...
 /***************************************************************************/
 
 typedef struct {
-     nlopt_func f; void *f_data;
-     int m, mm; nlopt_constraint *fc;
-     int p, pp; nlopt_constraint *h;
-     double rho, *lambda, *mu;
-     double *restmp, *gradtmp;
-     nlopt_stopping *stop;
+  nlopt_func f;
```
stevengj: All of these extraneous formatting and whitespace changes make this PR very difficult to review.

astamm: Oh my. So sorry about that. Automatic formatting in my IDE that I forgot to turn off. I can try to revert that.

astamm: I can't seem to separate well functional changes from formatting ones. What are the linting rules for nlopt? 4 spaces, no whitespaces, curly open braces on new line?

astamm: I see different conventions in different source files. Am I wrong? If we can agree on conventions, I can reformat everything accordingly.

astamm: Something like that should match the convention used in most files:

```
---
# Clang-format configuration derived from the observed style in bobyqa.c and
# other NLopt C source files under src/libs/.
#
# Key characteristics:
#   - 4-space indentation (spaces in function bodies; tabs only for parameter-
#     list continuation alignment are left as-is via UseTab: ForIndentation)
#   - Linux brace style: function-definition opening brace on its own line,
#     control-flow opening brace on the same line
#   - No column-limit enforcement (f2c-generated code has very long lines)
#   - Pointer qualifier attached to the variable name  (double *x)
#   - No include sorting (includes are grouped intentionally)

Language:        Cpp
BasedOnStyle:    LLVM

# Indentation
IndentWidth:               4
TabWidth:                  8
UseTab:                    ForIndentation
ContinuationIndentWidth:   4

# Braces — Linux style: function body brace on new line, rest on same line
BreakBeforeBraces:         Linux

# Pointer / reference qualifier attached to the variable name
PointerAlignment:          Right

# Line length — disabled; legacy f2c code has lines well beyond 80 columns
ColumnLimit:               0

# Short constructs
AllowShortFunctionsOnASingleLine:       None
AllowShortIfStatementsOnASingleLine:    Never
AllowShortLoopsOnASingleLine:           false
AllowShortBlocksOnASingleLine:          Empty

# Alignment — keep off to avoid churning legacy code
AlignConsecutiveAssignments:            None
AlignConsecutiveDeclarations:           None
AlignTrailingComments:
  Kind:                    Never

# Spaces
SpaceAfterCStyleCast:                   false
SpaceBeforeParens:                      ControlStatements
SpaceInEmptyParentheses:                false

# Includes — do not sort; order is intentional in NLopt sources
SortIncludes:              Never
IncludeBlocks:             Preserve
```

Should I apply this everywhere? Would that suit you?

It can be stored in a `.clang-format` hidden file at the root of the project so that we always follow these conventions. Let me know if this is something you would consider.

astamm: However, whitespaces after commas are automatically (in the sense hardcoded) emitted by clang formatter.

Also, the formatting style differs from the one adopted by the `f2c` tool that has been used for translating Fortran code. Do you have strong formatting rules for the nlopt project? Defining such rules in a file like `.clang-format` as I propose would avoid future formatting issues we are facing here.

aitap: Here's what can be tried to avoid digging the problem even deeper:
* Create a new branch from `git merge-base master r-compatibility`, let's call it `master-formatted`
* Format it using the closest `clang-format` settings you can find
* Format `r-compatibility` using the same settings
* Compute the diff from `master-formatted` to formatted `r-compatibility`
* Apply it on top of the original `git merge-base master r-compatibility`, solve the few remaining conflicts, restore the original code style, split into logical commits

This is probably easier than disentangling the formatting changes from semantic changes in the current branch.

astamm: On it, thanks for the tips!

astamm: Once found closest `clang-format` settings, don't we want to stick with it?
Do you already have strong formatting conventions for nlopt?

aitap: That's not for me to decide. I'm sorry I did not make it clear enough that I'm only giving advice (motivated by what I think is best for NLopt and the R community). In my experience, PRs that reformat the entire code base don't go well.

stevengj:
Don't use `clang-format` or do *any* wholesale reformatting.  (It could be quite reasonable to reformat nlopt using clang-format in a *separate* PR.)  Right now it's not any consistent formatting for historical reasons, but addressing that is orthogonal to the current PR.

For this PR, use a different IDE if you have to — get an editor where you can make localized changes without doing wholesale reformatting.  I'm simply not going to a merge a PR that includes 10,000 lines of extraneous formatting changes.

astamm:
On it, on it.

astamm:
We should be good now.

astamm:
Btw, this PR also addresses #604 #611 #612 that I filed early on.

src/algs/bobyqa/bobyqa.c
```diff
@@ -41,15 +43,28 @@ static void update_(int *n, int *npt, double *bmat,
...
+        bmat = _v - bmat_offset;
+    }
+    {
+        double *volatile _v = vlag;
+        vlag = _v - 1;
+    }
+    {
+        double *volatile _v = w;
+        w = _v - 1;
+    }
```
stevengj, line 67: This change seems unrelated to the topic of the PR?

astamm: You'll find this pattern repeatedly. It fixes GCC undefined behavior sanitizer issues that were spotted when I copy-pasted nlopt sources in the nloptrbundled package so I provided the fix. I believe it should improve compatibility with any downstream language that would like to provide a binding of nlopt. Hence, its presence here. The name of the PR though could be changed to better reflect what the provided solution actually does wrt the original intent.

astamm:
Please let me know if I can do something else to get this PR through. It would be really helpful for downstream packages in other languages that provide bindings to nlopt. Thanks!

astamm:
Can I do more to get this PR going on? Please let me know @stevengj. Thanks a lot!

astamm:
Any update on this? What else can I do to move this PR forward please? I'd like to get a new release of nloptr out soon on CRAN with this improvement. Thanks.

astamm:
Please @stevengj @aitap, let me know if I can do something else to get this PR through. It would be really helpful for downstream packages in other languages that provide bindings to nlopt. Thanks!

PR #612, opened 2025-04-23
warning: null format string in stop.c

astamm:
`gcc-ubsan` and `gcc-asan` show

```bash
warning: null format string
```

occurring in `stop.c` specifically in function `nlopt_vsprintf()`.

According to https://gcc.gnu.org/bugzilla/show_bug.cgi?id=116834, the fix is to first check that the pointer to `format` is not null and abort if it is before calling `vsnprintf()`.


astamm:
It actually happens at build time of *nlopt* with gcc-asan, prior to using its function in **nloptr**. Other 3 warnings seem to be raised: see details in [00install.out.txt](https://github.com/user-attachments/files/19875936/00install.out.txt).

PR #615, opened 2025-05-04
direct: diff from SciPy

lucascolley:
This draft PR shows the diff resulting from replacing the files of the direct algorithm here with what we have in SciPy. It looks like there is a mixture of (many) small formatting things, and (some) substantive changes.

Would there be interest into merging these sources into 'one true source'? This seems desirable, to de-duplicate work, ensuring any fixes made here end up in SciPy, and any contributions to SciPy end up here. We could set up a more formal vendoring situation to facilitate this once the sources are matched.

It may be that some of the changes here seem problematic, and we would rather change things in SciPy to match what is already here. That is also on the table, let me know what you think!

Cross reference https://github.com/scipy/scipy/issues/21232

### DIFF
--- a/src/algs/direct/DIRect.c
+++ b/src/algs/direct/DIRect.c
@@ -1,10 +1,10 @@
 /* DIRect-transp.f -- translated by f2c (version 20050501).
-   
-   f2c output hand-cleaned by SGJ (August 2007). 
+
+   f2c output hand-cleaned by SGJ (August 2007).
 */
 
-#include <math.h>
 #include "direct-internal.h"
+#include <math.h>
 
 /* Common Block Declarations */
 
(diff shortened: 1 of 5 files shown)


lucascolley:
if there is interest in proceeding here, perhaps we could merge all of the trivial formatting changes, then take a closer look at how to deal with the reduced diff

src/algs/direct/DIRect.c
```diff
@@ -80,40 +78,91 @@
...
+    #define MAXMEMORY 1073741824
+    integer MAXFUNC = *maxf <= 0 ? 101000 : (*maxf + 1000 + *maxf / 2);
+    integer fixed_memory_dim = ((*n) * (sizeof(doublereal) + sizeof(integer)) +
+                                      (sizeof(doublereal) * 2 + sizeof(integer)));
+    MAXFUNC = MAXFUNC * fixed_memory_dim > MAXMEMORY ? MAXMEMORY/fixed_memory_dim : MAXFUNC;
+    c__ = (doublereal *) malloc(sizeof(doublereal) * (MAXFUNC * (*n)));
+    if (!(c__)) {
+        *ierror = -100;
+        goto cleanup;
+    }
```
dschmitz89, line 102: This part was added as the algorithm uses absurd amounts of memory for large number of function evaluations. The `MY_ALLOC` macro did not help, we got a segmentation fault. Could this be of interest to nlopt @stevengj ?

dschmitz89:
A lot of these change change functions from `void` to `PyObject`. This makes sense for scipy but probably not so much for a C library like nlopt. See above for one finding we had related to memory issues. I guess a better solution long term would be to wrap nlopt's reimplementation of DIRECT which uses trees instead of arrays to reduce memory usage.

dschmitz89:
CC @czgdp1807

Issue #616, opened 2025-05-22
Building octave bindings for octave >= 10 requires c++17 standard


badshah400:
### What happened?

Octave versions 10 and above require the C++ standard to be [at least c++17](https://octave.org/NEWS-10.html) for code to be built against it. Since nlopt enforces c++11 for its builds, building octave bindings fail when the latest version of octave (10.1.0) is used, as we do now on openSUSE Tumbleweed. The log file snippet pasted below shows the issue.

Upgrading the `CMAKE_CXX_STANDARD` to 17 from 11 fixes builds.

```log
[   30s] [ 77%] Building CXX object src/octave/CMakeFiles/nlopt_optimize.dir/nlopt_optimize-oct.cc.o
[   30s] cd /home/abuild/rpmbuild/BUILD/nlopt-2.10.0-build/python3.11_build/build/src/octave && /usr/bin/c++ -DNLOPT_DLL -Dnlopt_optimize_EXPORTS -I/usr/include/octave-10.1.0 -I/usr/include/octave-10.1.0/octave -I/home/abuild/rpmbuild/BUILD/nlopt-2.10.0-build/python3.11_build/build/src/octave -I/home/abuild/rpmbuild/BUILD/nlopt-2.10.0-build/python3.11_build/src/api -I/home/abuild/rpmbuild/BUILD/nlopt-2.10.0-build/python3.11_build/build -O2 -Wall -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2 -fstack-protector-strong -funwind-tables -fasynchronous-unwind-tables -fstack-clash-protection -Werror=return-type -flto=auto -g -O2 -g -DNDEBUG -std=gnu++11 -fPIC -MD -MT src/octave/CMakeFiles/nlopt_optimize.dir/nlopt_optimize-oct.cc.o -MF CMakeFiles/nlopt_optimize.dir/nlopt_optimize-oct.cc.o.d -o CMakeFiles/nlopt_optimize.dir/nlopt_optimize-oct.cc.o -c /home/abuild/rpmbuild/BUILD/nlopt-2.10.0-build/python3.11_build/src/octave/nlopt_optimize-oct.cc
[   30s] In file included from /usr/include/octave-10.1.0/octave/Array.h:37,
[   30s]                  from /usr/include/octave-10.1.0/octave/Array-util.h:31,
[   30s]                  from /usr/include/octave-10.1.0/octave/MSparse.h:31,
[...]

PR #618, opened 2025-06-08
Octave: Use c++ 17

jschueller:
see #616


opoplawski:
This patch does not appear to be working for me in Fedora rawhide with octave 10.1.0.  I still end up with `-std=gnu++11`.  I that version logic really correct?

PR #623, opened 2025-08-23
new CCSA and MMA parameters: inner_gradients and always_improve

### LINKED ISSUE #376: Discriminating inner and outer function calls for MMA
I've been using the MMA algorithm and so far it converges (albeit slowly) to a local minimum for my optimization problem. However, I have some questions about the implementation that should allow me to accelerate the implementation. 

1. I can calculate arbitrarily accurate gradients for my function of N parameters (and I have verified with finite differences that my gradients are correct), however, calculating the gradients takes ~N times as long as just calculating the objective value. The MMA paper in the documentation is quoted as "Each new inner iteration requires function values, but no derivatives, calculated at the optimal solution of the most recent subproblem." So is there a way to discriminate between inner function calls and outer function calls? I have observed that in SLSQP sometimes the objective function is called with a gradient vector of length 0. I assume that those are "inner calls," and can execute them more quickly. However, the MMA algorithm does not do this. Is there a solution here?

2. When I log the objective function minimization and observe how it converges with iteration I see that the objective does not decrease monotonically. As far as I can tell from the original MMA paper, this should not be the case. However, there are periods of monotonic convergence interspersed by periods of non-montonicity. Because I can't discriminate inner vs. outer function calls I can't tell whether the non-monotonicity is simply due to internal function calls, in which case I wouldn't have a problem. Are inner calls guaranteed to also be monotonically decreasing for MMA?

Also, are there any suggestions for how many parameters are feasible for SQSLP? MMA says 10^4-10^5. 

Thanks!

### LINKED ISSUE #484: Reduce number of gradient calculations in LD-MMA
The Svanberg MMA paper notes that for the CCSA algorithms described, gradients are only required in the outer iterations. "Each new inner iteration requires function values, but no derivatives."

However, it appears that the implementation of `LD-MMA` calculates a gradient in the inner as well as the outer iteration. I request that the implementation be updated to reduce gradient calculation.

I believe this is a two-line change: [This line](https://github.com/stevengj/nlopt/blob/e30e2b0415a99d9c9a11e841ef8432d0992376c4/src/algs/mma/mma.c#L292) could be changed to something like `fcur = f(n, xcur, NULL, f_data);`, and then after line 299 in the same file one could add the code `if (inner_done) { fcur = f(n, xcur, dfdx_cur, f_data); }`. 

This would duplicate objective calls once per outer iteration, but since gradient calculations tend to dominate run-time in objective function calls, there should be overall net savings whenever more than one inner iteration is used.

I regret not being able to try this out myself. I don't have C set up on my machine and have never coded in C, so I would be extremely slow at running tests (I'm using the Python API). Thanks for considering!

stevengj:
This PR adds two new algorithm-specific parameters to the CCSA and MMA algorithms, one to control whether gradients are requested on inner iterations, and one to control how aggressive the algorithm is about updating the current point (see the new docs).

The default behavior is not changed, in order to not hurt existing code that uses these algorithms.  For example, not computing the gradient on inner iterations actually hurts performance in the common case where it succeeds after a single inner iteration — since it has to then call the function *again* with gradients on the outer iteration — unless the caller optimizes their objective function (and constraints) to cache the current `x` point and not recompute the function (only compute the new gradient) if it is called twice in a row with the same `x`.

Closes #484.  Closes #376.


smartalecH (commented):
Thanks @stevengj 

Suppose we want to only perform one outer step per "stochastic sample" (and as many inner steps as needed to hit our convergence criteria.)

Is there an API to specify this behavior? In other words, can we specify the "max number of outer iterations"? We have a parameter for the max number of inner iterations, and _all_ algorithms can specify a max number of function calls (which in this case includes outer _and_ inner iterations). But I don't see a mechanism to specify a max number of outer iterations implicitly or explicitly.

stevengj:
Good idea.  It would be easy to add another algorithm parameter that makes the maxevals apply only to outer iterations.

Another option would be to count gradient evaluations an an eval, i.e. count function+gradient (outer iterations) as two evals and function-only (inner iterations) as one eval.   (If you were benchmarking two algorithms against one another you would probably want to do this.)

stevengj:
This is proving useful enough to merge as-is, I think.

PR #646, opened 2026-08-21
Create SECURITY.md for security policy

jet-byye:
Added a security policy document outlining supported versions and vulnerability reporting.


jschueller:
I wouldnt consider nlopt security-sensitive

stevengj:
Yes, I don't think that this makes sense here.

